PR #5, opened 2024-03-09
Improve Dockerfile

AlexGluck:
1. Union argument layers and copy source code.
2. Changed layers order final image for grow cache hit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,17 @@
 # Build the manager binary
 FROM golang:1.21 AS builder
-ARG TARGETOS
-ARG TARGETARCH
+ARG TARGETOS TARGETARCH
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
+COPY go.mod go.sum ./
+
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
 # Copy the go source
-COPY cmd/main.go cmd/main.go
-COPY api/ api/
-COPY internal/controller/ internal/controller/
+COPY cmd/main.go api/ internal/controller/ ./
 
 # Build
 # the GOARCH has not a default value to allow the binary be built according to the host where the command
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -26,8 +23,7 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
```
sircthulhu, line 25: Should we set specific version of distroless image?

sergeyshevch: AFAIK It doesn't specified mos of the times

Dockerfile
```diff
@@ -1,20 +1,17 @@
...
+
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
 # Copy the go source
-COPY cmd/main.go cmd/main.go
-COPY api/ api/
-COPY internal/controller/ internal/controller/
+COPY cmd/main.go api/ internal/controller/ ./
```
sircthulhu, line 14: This line is not working correctly.
```
 > [builder 6/7] RUN ls -la && exit 1:                                                                                                                                                      
0.082 total 32                                                                                                                                                                              
0.082 drwxr-xr-x 1 root root    30 Mar 11 07:41 .                                                                                                                                           
0.082 drwxr-xr-x 1 root root     6 Mar 11 08:01 ..                                                                                                                                          
0.082 -rw-r--r-- 1 root root     0 Mar 11 07:41 .gitkeep                                                                                                                                    
0.082 -rw-r--r-- 1 root root  3175 Mar 10 07:46 go.mod
0.082 -rw-r--r-- 1 root root 18606 Mar 10 07:46 go.sum
0.082 -rw-r--r-- 1 root root  4651 Mar 10 07:46 main.go
```

sircthulhu (changes requested):

sircthulhu:
Please, rebase branch on top of `main` branch for `api/` and `internal/controller` directories to appear, so that build won't fail on not found error

apten-fors (approved):
lgtm

Issue #13, opened 2024-03-12
Write basic etcd-operator logic


kvaps:
We agreed on the folowing logic:

- Create configmap with env `ETCD_INITIAL_CLUSTER_STATE=new`
- Create StatefulSet with needed amount of replicas and `envFrom: <configmap>`, and `podManagementPolicy: Parallel`
- Wait unitl all replicas become ready
- Update configmap to  `ETCD_INITIAL_CLUSTER_STATE=existing`
- Consider that cluster as bootstrapped

currently locked by https://github.com/aenix-io/etcd-operator/issues/9

apten-fors:
Don't forget about persistence:
`spec:
  persistentVolumeClaimRetentionPolicy:
    whenDeleted: Retain
    whenScaled: Retain
  podManagementPolicy: Parallel
`

kvaps:
> `podManagementPolicy: Parallel`

I think this policy should not be controlled by user, as it is part of operator logic

> ```
>   persistentVolumeClaimRetentionPolicy:
>     whenDeleted: Retain
>     whenScaled: Retain
> ```

Do you have any examples how are other operator define this spec?
Let's discuss it at the next community meeting.

PR #21, opened 2024-03-14
Implementation of basic controller logic

sircthulhu:
Controller now can successfully bootstrap etcd cluster :)

Controller creates Service, ConfigMap and StatefulSet based on EtcdCluster CR
Only EmptyDir storage is supported
Initialized status is being set and updated

fixes https://github.com/aenix-io/etcd-operator/issues/13

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -71,7 +71,7 @@ spec:
         args:
         - --leader-elect
         image: controller:latest
-        imagePullPolicy: Never
+        imagePullPolicy: IfNotPresent
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -71,6 +71,7 @@ spec:
         args:
         - --leader-elect
         image: controller:latest
+        imagePullPolicy: Never
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -70,7 +70,7 @@ spec:
         - /manager
         args:
         - --leader-elect
-        image: controller:latest
+        image: ghcr.io/aenix-io/etcd-operator:latest
         imagePullPolicy: IfNotPresent
         name: manager
         securityContext:
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -63,7 +63,7 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return reconcile.Result{}, err
 	}
 	// If object is being deleted, skipping reconciliation
-	if instance.DeletionTimestamp != nil && !instance.DeletionTimestamp.IsZero() {
+	if !instance.DeletionTimestamp.IsZero() {
 		return reconcile.Result{}, nil
 	}
 
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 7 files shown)


config/manager/manager.yaml
```diff
@@ -71,6 +71,7 @@ spec:
         args:
         - --leader-elect
         image: controller:latest
+        imagePullPolicy: Never
```
AlexGluck, line 74: Suggest use full image path `ghcr.io/aenix-io/controller:latest` and `imagePullPolicy: IfNotPresent` for reuse yaml in static manifests and helm.

internal/controller/etcdcluster_controller.go
```diff
@@ -36,24 +44,349 @@ type EtcdClusterReconciler struct {
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.V(2).Info("object not found", "namespaced_name", req.NamespacedName)
+			return ctrl.Result{}, nil
+		}
+		// Error retrieving object, requeue
+		return reconcile.Result{}, err
+	}
+	// If object is being deleted, skipping reconciliation
+	if instance.DeletionTimestamp != nil && !instance.DeletionTimestamp.IsZero() {
```
aobort, line 66: Comparison with nil is redundant, due to `IsZero()` method already includes this check.

internal/controller/etcdcluster_controller.go
```diff
@@ -36,24 +44,349 @@ type EtcdClusterReconciler struct {
...
+	notFound := false
+	if errors.IsNotFound(err) {
+		notFound = true
+		// prepare initial cluster members
+		initialCluster := ""
+		for i := uint(0); i < cluster.Spec.Replicas; i++ {
+			if i > 0 {
+				initialCluster += ","
+			}
+			initialCluster += fmt.Sprintf("%s-%d=https://%s-%d.%s.%s.svc.cluster.local:2380",
```
kvaps, line 236: Please do not hardcode `cluster.local` suffixes as many clusters may have different ones.

It's possible to access services without them.

internal/controller/etcdcluster_controller.go
```diff
@@ -36,24 +44,349 @@ type EtcdClusterReconciler struct {
...
+									"--listen-peer-urls=https://0.0.0.0:2380",
+									// for first version disable TLS for client access
+									"--listen-client-urls=http://0.0.0.0:2379",
+									"--initial-advertise-peer-urls=https://$(POD_NAME)." + cluster.Name + ".$(POD_NAMESPACE).svc.cluster.local:2380",
+									"--data-dir=/var/run/etcd/default.etcd",
+									"--initial-cluster=" + initialCluster,
+									fmt.Sprintf("--initial-cluster-token=%s-%s", cluster.Name, cluster.Namespace),
+									"--auto-tls",
+									"--peer-auto-tls",
+									"--advertise-client-urls=http://$(POD_NAME)." + cluster.Name + ".$(POD_NAMESPACE).svc.cluster.local:2379",
```
kvaps, line 284: Please do not hardcode `cluster.local` suffixes as many clusters may have different ones.

It's possible to access services without them.

internal/controller/etcdcluster_controller.go
```diff
@@ -36,24 +44,349 @@ type EtcdClusterReconciler struct {
...
+				"app.kubernetes.io/managed-by": "etcd-operator",
+			},
+			PublishNotReadyAddresses: true,
+		},
+	}
+	if err = ctrl.SetControllerReference(cluster, svc, r.Scheme); err != nil {
+		return fmt.Errorf("cannot set controller reference: %w", err)
+	}
+	if err = r.Create(ctx, svc); err != nil {
+		return fmt.Errorf("cannot create cluster state configmap: %w", err)
```
Fhtgn, line 178: There should be service instead of configmap

config/manager/manager.yaml
kvaps: It should be `ghcr.io/aenix-io/etcd-operator/controller:latest` or `ghcr.io/aenix-io/etcd-operator:latest`

kvaps: Or `ghcr.io/aenix-io/etcd-operator/etcd-operator:latest`

@hiddenmarten could you comment on this, please?

internal/controller/etcdcluster_controller.go
Fhtgn, line 236: Don't use cluster.local, it's may be different in some clusters

internal/controller/etcdcluster_controller.go
Fhtgn, line 284: Don't use cluster.local, it's may be different in some clusters

config/manager/manager.yaml
hiddenmarten: @kvaps 
Sure, I think it's better to use a simple structure of image tag, that will correspond with a GitHub repo path (It helps to avoid confusion):
`{registry}/{organisation}/{application}:{versionBySemVer}`

So, as a result:
`ghcr.io/aenix-io/etcd-operator:v0.0.0`

kvaps (approved):

Fhtgn (approved):

PR #26, opened 2024-03-15
Implementation of Ready status condition

### LINKED ISSUE #24: Implement Ready status condition
# In general
We need to implement checking quorum status of an initialized cluster and update `Ready` status condition in accordance. After initializing cluster and making sure pods found each other and formed a cluster, controller must update cluster state configmap to set cluster state `existing` (from `new`)

# Design proposal
When cluster is initialized, we should check if:
1. StatefulSet is `Ready`
2. Connect to any etcd and check if quorum is fine
- if everything is fine, set status condition Ready and update value in cluster state configmap to `existing`.
- If some pods in statefulset aren't ready but quorum is fine, set Ready, but mark cluster as degraded in Reason field.
- If pods in sts aren't ready and quorum is lost, mark cluster as not Ready

If configmap already has `existing` state, do not change it anymore, as cluster should be already initialized

sircthulhu:
Update cluster Ready status according to StatefulSet status and update ConfigMap cluster state to `existing` after first time STS is ready
fixes #24

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -21,6 +21,9 @@ import (
 	"fmt"
 	"slices"
 
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -437,7 +440,7 @@ func (r *EtcdClusterReconciler) updateClusterState(cluster *etcdaenixiov1alpha1.
 // SetupWithManager sets up the controller with the Manager.
 func (r *EtcdClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&etcdaenixiov1alpha1.EtcdCluster{}).
+		For(&etcdaenixiov1alpha1.EtcdCluster{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&appsv1.StatefulSet{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.Service{}).
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -71,10 +71,10 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if len(instance.Status.Conditions) == 0 {
 		instance.Status.Conditions = append(instance.Status.Conditions, metav1.Condition{
 			Type:               etcdaenixiov1alpha1.EtcdConditionInitialized,
-			Status:             "False",
+			Status:             metav1.ConditionFalse,
 			ObservedGeneration: instance.Generation,
 			LastTransitionTime: metav1.Now(),
-			Reason:             "InitializationStarted",
+			Reason:             string(etcdaenixiov1alpha1.EtcdCondTypeInitStarted),
 			Message:            "Cluster initialization has started",
 		})
 	}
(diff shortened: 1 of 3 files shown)


hiddenmarten (approved):
lgtm

internal/controller/etcdcluster_controller.go
```diff
@@ -84,41 +84,54 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		instance.Status.Conditions[initIdx].Status = "True"
-		instance.Status.Conditions[initIdx].LastTransitionTime = metav1.Now()
-		instance.Status.Conditions[initIdx].Reason = "InitializationComplete"
-		instance.Status.Conditions[initIdx].Message = "Cluster initialization is complete"
-	} else {
-		instance.Status.Conditions = append(instance.Status.Conditions, metav1.Condition{
-			Type:               etcdaenixiov1alpha1.EtcdConditionInitialized,
+	r.updateClusterState(instance, metav1.Condition{
+		Type:               etcdaenixiov1alpha1.EtcdConditionInitialized,
+		Status:             "True",
```
Uburro, line 104: metav1.ConditionTrue

internal/controller/etcdcluster_controller.go
```diff
@@ -84,41 +84,54 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		LastTransitionTime: metav1.Now(),
+		Reason:             "InitializationComplete",
```
Uburro, line 106: IMHO better us a a const in an api package 
for instance:
/api/v1alpha1
```
type ETCDConditionType string
const (
	ETCDCondTypeInitComplete    ETCDConditionType = "InitializationComplete"
)
```

Reason: string(etcdaenixiov1alpha1.ETCDCondTypeInitComplete)

internal/controller/etcdcluster_controller.go
```diff
@@ -324,7 +345,7 @@ func (r *EtcdClusterReconciler) ensureClusterStatefulSet(
 								LivenessProbe: &corev1.Probe{
```
Uburro, line 345: let's add [StartupProbe](https://github.com/kubernetes/kubernetes/blob/89b1db79d75e367f567ea06c0528ef4b8f3fecb4/cmd/kubeadm/app/phases/etcd/local.go#L228) too 
with /readyz??serializable=false

Uburro: https://github.com/kubernetes/kubernetes/commit/2829fc03e34b021adb83e8dff8d404ac981a49a2

internal/controller/etcdcluster_controller.go
```diff
@@ -387,9 +408,26 @@ func (r *EtcdClusterReconciler) getClusterStateConfigMapName(cluster *etcdaenixi
...
+		state.ObservedGeneration = cluster.Generation
+		cluster.Status.Conditions = append(cluster.Status.Conditions, state)
+	}
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *EtcdClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&etcdaenixiov1alpha1.EtcdCluster{}).
+		Owns(&appsv1.StatefulSet{}).
```
sergeyshevch, line 431: Let's also add Owns for ConfigMap/Service.

I guess we want to reconcile cluster if someone accidentally delete some of child resources

internal/controller/etcdcluster_controller.go
```diff
@@ -387,9 +418,28 @@ func (r *EtcdClusterReconciler) getClusterStateConfigMapName(cluster *etcdaenixi
...
+	} else {
+		state.ObservedGeneration = cluster.Generation
+		cluster.Status.Conditions = append(cluster.Status.Conditions, state)
+	}
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *EtcdClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&etcdaenixiov1alpha1.EtcdCluster{}).
```
Uburro, line 440: For(&etcdaenixiov1alpha1.EtcdCluster{},
builder.WithPredicates(predicate.GenerationChangedPredicate{})).

Uburro (approved):

PR #41, opened 2024-03-19
enh: Update status in reconcile loop, not in defer

### LINKED ISSUE #36: Move EtcdCluster update call out of defer function
In case an error occur on EtcdCluster resource update (https://github.com/aenix-io/etcd-operator/blob/main/internal/controller/etcdcluster_controller.go#L84-L88):

```go
...
defer func() {
	if err := r.Status().Update(ctx, instance); err != nil && !errors.IsConflict(err) {
		logger.Error(err, "unable to update cluster")
	}
}()
...
return ctrl.Result{}, nil
```

it will not be returned and request will not be requeued. Therefore update call should be moved out of defer function.

sircthulhu:
In case error happens during status update request, reconcile will not be requeued. So, this call should be moved ouside `defer` function
fixes #36

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -81,11 +81,6 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			Message:            "Cluster initialization has started",
 		})
 	}
-	defer func() {
-		if err := r.Status().Update(ctx, instance); err != nil && !errors.IsConflict(err) {
-			logger.Error(err, "unable to update cluster")
-		}
-	}()
 
 	// check sts condition
 	isClusterReady := false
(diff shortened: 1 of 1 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -99,7 +94,11 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	if err := r.ensureClusterObjects(ctx, instance, isClusterReady); err != nil {
-		return ctrl.Result{}, fmt.Errorf("cannot create Cluster auxiliary objects: %w", err)
+		res, err2 := r.updateStatus(ctx, instance)
```
sergeyshevch, line 97: Maybe name it as statusErr?

sergeyshevch: Or as another option you can create function updateStatusOnErr and just always return this function in case of error. This function will wrap error and update status internally

sergeyshevch (changes requested):

internal/controller/etcdcluster_controller.go
```diff
@@ -99,7 +94,8 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	if err := r.ensureClusterObjects(ctx, instance, isClusterReady); err != nil {
-		return ctrl.Result{}, fmt.Errorf("cannot create Cluster auxiliary objects: %w", err)
+		logger.Error(err, "cannot create Cluster auxiliary objects")
+		return r.updateStatusOnErr(ctx, instance, fmt.Errorf("cannot create Cluster auxiliary objects: %w", err))
```
sergeyshevch, line 98: We will duplicate this error wrapping. We wrap it with text here and in L480

sergeyshevch (approved):

PR #44, opened 2024-03-19
feat: Create ClusterIP Service for clients access to cluster

sircthulhu:
in order to access cluster from clients using one IP we decided to use ClusterIP Service

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -144,6 +144,10 @@ func (r *EtcdClusterReconciler) ensureClusterObjects(
 	if err := r.ensureClusterStatefulSet(ctx, cluster); err != nil {
 		return err
 	}
+	// 3. create or update ClusterIP Service
+	if err := r.ensureClusterClientService(ctx, cluster); err != nil {
+		return err
+	}
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


sergeyshevch (approved):
LGTM

internal/controller/etcdcluster_controller.go
```diff
@@ -144,6 +144,10 @@ func (r *EtcdClusterReconciler) ensureClusterObjects(
 	if err := r.ensureClusterStatefulSet(ctx, cluster); err != nil {
 		return err
 	}
+	// 3. create or update ClusterIP Service
+	if err := r.ensureClusterClientService(ctx, cluster); err != nil {
+		return err
```
Kirill-Garbar, line 149: Do I get it right that return err will make operator exit? Or will it just rise log message with error?

internal/controller/etcdcluster_controller.go
```diff
@@ -196,6 +200,52 @@ func (r *EtcdClusterReconciler) ensureClusterService(ctx context.Context, cluste
...
+			Ports: []corev1.ServicePort{
+				{Name: "client", TargetPort: intstr.FromInt32(2379), Port: 2379, Protocol: corev1.ProtocolTCP},
+			},
+			Type: corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{
+				"app.kubernetes.io/name":       "etcd",
+				"app.kubernetes.io/instance":   cluster.Name,
+				"app.kubernetes.io/managed-by": "etcd-operator",
+			},
+			PublishNotReadyAddresses: false,
```
gvencadze, line 237: probably we don't need to set `false` value for `PublishNotReadyAddresses` parameter explicitly , coz it's false by default.

internal/controller/etcdcluster_controller.go
sircthulhu: If error is returned, reconciler will exit `reconcile` method with error and set CR condition to type InitializationStarted, status false

internal/controller/etcdcluster_controller_test.go
```diff
@@ -110,12 +110,21 @@ var _ = Describe("EtcdCluster Controller", func() {
...
 			Expect(err).NotTo(HaveOccurred(), "cluster statefulset should exist")
+			// check that Service is created
+			svc = &v1.Service{}
+			clientSvcName := types.NamespacedName{
+				Namespace: typeNamespacedName.Namespace,
+				Name:      controllerReconciler.getClientServiceName(etcdcluster),
+			}
+			err = k8sClient.Get(ctx, clientSvcName, svc)
+			Expect(err).NotTo(HaveOccurred(), "cluster client Service should exist")
+			Expect(svc.Spec.ClusterIP).NotTo(Equal("None"), "cluster client Service should NOT be headless")
```
Kirill-Garbar, line 127: Does it make sense to check correct ports and label selectors to pods?

Kirill-Garbar (approved):

internal/controller/etcdcluster_controller_test.go
sircthulhu: What about ports, I don't think so. It'd better to check in e2e tests

Writing tests for label selectors won't be good enough neither as we can only test the perfect case - all service label selector are included in pods` labels, but this does not protect us from collisions

PR #47, opened 2024-03-19
[43] Fix go vet hags on first run

hiddenmarten:


sircthulhu:
@hiddenmarten 
Could you please explain why this change will fix this error?

hiddenmarten:
@sircthulhu @sergeyshevch 

By default, pre-commit operates using multiple threads in parallel. However, this flag alters its behavior to use only one thread. I've tested this adjustment on my local setup and it helps.

AlexGluck (dismissed):
I think we need change behavior instead of broke parallel work of linters (this worsens performance).

genestack-okunitsyn:
> I think we need change behavior instead of broke parallel work of linters (this worsens performance).

Do you have any suggestions?

hiddenmarten:
Probably I gave you an incorrect explanation about the behavior of pre-commit. 
I meant processes, not threads.

By default, pre-commit runs a bunch of processes
```
make-vet.................................................................Passed
- hook id: make-vet
- duration: 0.95s

go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
go vet ./...
```
however, with this option pre-commit runs `go vet ./...` only once, which means parallelization (as well as performance) of the linter itself isn't affected:

```
make-vet.................................................................Passed
- hook id: make-vet
- duration: 0.33s

go vet ./...

```

The root issue is about concurrent `go get .` calls triggered by `go vet ./...`.

gecube (approved):

AlexGluck (approved):
lgtm

PR #57, opened 2024-03-20
Add validation minimum and change type to int32 for compatibility

### LINKED ISSUE #51: Move defaulting from webhooks to CRD
Consider replacing defaulting webhook with native CRD validation:

https://github.com/aenix-io/etcd-operator/blob/94c538dbc1e500bed42f5a0ff21edb5324b2690f/api/v1alpha1/etcdcluster_webhook.go#L45-L47

https://github.com/aenix-io/etcd-operator/blob/94c538dbc1e500bed42f5a0ff21edb5324b2690f/api/v1alpha1/etcdcluster_types.go#L33

sircthulhu:
- `Replicas` field type changed to *int32 according to k8s api guidelines
- remove unnecessary validation (everything is validated using OpenAPI scheme)
- allow creation of cluster with 0 replicas

fixes #51


api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -42,9 +42,6 @@ var _ webhook.Defaulter = &EtcdCluster{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *EtcdCluster) Default() {
 	etcdclusterlog.Info("default", "name", r.Name)
-	if r.Spec.Replicas == 0 {
-		r.Spec.Replicas = 3
-	}
 	if r.Spec.Storage.Size.IsZero() {
 		r.Spec.Storage.Size = resource.MustParse("4Gi")
 	}
```
aobort, line 47: `// +kubebuiler:default:="4Gi"` can be also added to type definition instead of webhook

sircthulhu: It is not possible for complex structs to set default value. The only way to define default value for `resource.Quantity` is by code.

I also checked your suggestion, it's not working, it remains zero.

Actually, we can leave it zero (without any defaults) and it'll be up to user to fill it

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -57,6 +54,14 @@ var _ webhook.Validator = &EtcdCluster{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *EtcdCluster) ValidateCreate() (admission.Warnings, error) {
 	etcdclusterlog.Info("validate create", "name", r.Name)
+	//var allErrors field.ErrorList
```
sergeyshevch, line 57: Maybe we should remove this commented code

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -68,10 +73,16 @@ func (r *EtcdCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, er
 		warnings = append(warnings, "cluster resize is not currently supported")
 	}
 
-	if len(warnings) > 0 {
-		return warnings, nil
-	}
-	return nil, nil
+	//var allErrors field.ErrorList
```
sergeyshevch, line 76: Maybe we should remove this commented code

api/v1alpha1/etcdcluster_webhook_test.go
```diff
@@ -28,40 +29,57 @@ var _ = Describe("EtcdCluster Webhook", func() {
...
 			etcdCluster.Default()
-			gomega.Expect(etcdCluster.Spec.Replicas).To(gomega.Equal(uint(5)))
+			gomega.Expect(*etcdCluster.Spec.Replicas).To(gomega.Equal(int32(5)))
 			gomega.Expect(etcdCluster.Spec.Storage.Size).To(gomega.Equal(resource.MustParse("10Gi")))
 		})
 	})
 
-	// Not yet applicable as currently all fields are optional.
+	Context("When creating EtcdCluster under Validating Webhook", func() {
+		//It("Should deny if replicas is negative", func() {
```
sergeyshevch, line 53: Maybe we should remove this commented code

sergeyshevch (approved):

api/v1alpha1/etcdcluster_webhook.go
sircthulhu: Checked one more time using local Kind cluster. Value remains zero if I use marker. So, I think we have two options
- leave it zero by default
- fill it using defaulting webhook

gecube (approved):

PR #72, opened 2024-03-22
feat: Implement podSpec generation

### LINKED ISSUE #69: Implement PodSpec for EtcdCluster CRD
Implement podspec according to proposal #62

sergeyshevch:
Fixes #69

### DIFF
--- a/internal/controller/factory/builders.go
+++ b/internal/controller/factory/builders.go
@@ -40,7 +40,7 @@ func reconcileSTS(ctx context.Context, rclient client.Client, crdName string, st
 		}
 		return fmt.Errorf("cannot get existing statefulset: %s, for crd_object: %s, err: %w", sts.Name, crdName, err)
 	}
-	sts.Annotations = labels.Merge(sts.Annotations, sts.Annotations)
+	sts.Annotations = labels.Merge(currentSts.Annotations, sts.Annotations)
 	if sts.ResourceVersion != "" {
 		sts.ResourceVersion = currentSts.ResourceVersion
 	}


sircthulhu (changes requested):
Good job! Just fix update validator and add license header to each go file please

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -79,7 +79,7 @@ func (r *EtcdCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, er
 	}
 
 	var allErrors field.ErrorList
-	if oldCluster.Spec.Storage.EmptyDir != r.Spec.Storage.EmptyDir {
+	if oldCluster.Spec.Storage.EmptyDir.String() != r.Spec.Storage.EmptyDir.String() {
```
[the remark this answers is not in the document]
sergeyshevch: Yes. This just fixes comparison of pointers. It always will be false and prevent any changes in CR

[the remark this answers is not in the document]
sergeyshevch: Other logic is same with your version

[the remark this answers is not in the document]
sircthulhu: Doesn't it reject changes to empty dir size?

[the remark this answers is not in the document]
sergeyshevch: .String() Return all struct value. Not only size. Logic remains unchanged, but yes. It also block any changes in empty dir size but we need to address it in other PR

sergeyshevch:
> add license header to each go file please
Fixed

internal/controller/factory/suite_test.go
sircthulhu, line 1: I think this file should be in `controller` directory

sircthulhu (approved):

internal/controller/factory/suite_test.go
sergeyshevch: There are both. It's different suites that perform different initialization AFAIK

sergeyshevch: Should be located in both packages. It create test client that later user in test suites

internal/controller/factory/builders.go
```diff
@@ -0,0 +1,49 @@
...
+	currentSts := &appsv1.StatefulSet{}
+	err := rclient.Get(ctx, types.NamespacedName{Namespace: sts.Namespace, Name: sts.Name}, currentSts)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.V(2).Info("creating new statefulset", "sts_name", sts.Name, "crd_object", crdName)
+			return rclient.Create(ctx, sts)
+		}
+		return fmt.Errorf("cannot get existing statefulset: %s, for crd_object: %s, err: %w", sts.Name, crdName, err)
+	}
+	sts.Annotations = labels.Merge(sts.Annotations, sts.Annotations)
```
Kirill-Garbar, line 43: Why do you merge sts.Annotations with themselves?

sergeyshevch: Typo
Should merge from current sts

Kirill-Garbar: If I got it right, it should be merged with currentSts

internal/controller/factory/builders.go
```diff
@@ -0,0 +1,49 @@
...
+	if sts.ResourceVersion != "" {
+		sts.ResourceVersion = currentSts.ResourceVersion
```
Kirill-Garbar, line 45: sts is what we build, currentSts - what we got from the cluster.

Is it just a measure to protect ourselves from accidential change of resource version outside this function?
Will it be bad if somebody changes sts manually, version is update and after that we return previous version in our code? Classic race condition.

If I got it right and it is just a measure to protect ouselves, maybe we cal make it null and do not pass it at all?

internal/controller/factory/statefulset.go
```diff
@@ -0,0 +1,231 @@
...
+								PeriodSeconds:       5,
+							},
+							ReadinessProbe: &corev1.Probe{
+								ProbeHandler: corev1.ProbeHandler{
+									HTTPGet: &corev1.HTTPGetAction{
+										Path: "/readyz",
+										Port: intstr.FromInt32(2379),
+									},
+								},
+								InitialDelaySeconds: 5,
```
Kirill-Garbar, line 155: Should we make probe settings customizable as well? 
Not for this MR. I've just thought that probe parameters could depend on infrastructure settings, volume and network latency.

internal/controller/factory/builders.go
sergeyshevch: We apply upgrade so it always should be resourceVersion from current STS

If there will be a race, we will get an error on upgrade call

internal/controller/factory/statefulset.go
sergeyshevch: This behaviour isn't changed in this PR. I like this idea but let's make this in follow-up PR

internal/controller/factory/builders.go
Kirill-Garbar: I have not developed for k8s a lot, but from my other experience resourceVersion is a value that should be changed only by k8s, not by etcd or any other custom operator. This code makes an attempt to change this value. My proposal is to avoid touching it at all. Just not to pass resource version. And if it was passed from another function => make it null in the object that will be passed to k8s api.

Anyway I will resolve this comment and in case of disagreement will find recommendations from k8s if there are any.

Kirill-Garbar (approved):

internal/controller/factory/statefulset.go
sergeyshevch: Follow up in #82

PR #84, opened 2024-03-25
feat: Refactor resource generation

### LINKED ISSUE #55: CreateOrUpdate cluster auxiliary entities
On cluster bootstrap we create 2 Service, ConfigMap and StatefulSet. After they are created, they won't be updated unless they're deleted by user. We should fix it and use `CreateOrUpdate` syntax

sergeyshevch:
Closes #55

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -132,10 +131,10 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 func (r *EtcdClusterReconciler) ensureClusterObjects(
 	ctx context.Context, cluster *etcdaenixiov1alpha1.EtcdCluster, isClusterInitialized bool) error {
 	// 1. create or update configmap <name>-cluster-state
-	if err := r.ensureClusterStateConfigMap(ctx, cluster, isClusterInitialized); err != nil {
+	if err := factory.CreateOrUpdateClusterStateConfigMap(ctx, cluster, isClusterInitialized, r.Client, r.Scheme); err != nil {
 		return err
 	}
-	if err := r.ensureClusterService(ctx, cluster); err != nil {
+	if err := factory.CreateOrUpdateClusterService(ctx, cluster, r.Client, r.Scheme); err != nil {
 		return err
 	}
 	// 2. create or update statefulset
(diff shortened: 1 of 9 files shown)


sircthulhu (approved):
I don't really like code style with functions of 4+ parameters but OK

internal/controller/factory/builders.go
```diff
@@ -47,3 +48,42 @@ func reconcileSTS(ctx context.Context, rclient client.Client, crdName string, st
...
+		return fmt.Errorf("cannot get existing configMap: %s, for crd_object: %s, err: %w", configMap.Name, crdName, err)
+	}
+	configMap.Annotations = labels.Merge(currentConfigMap.Annotations, configMap.Annotations)
+	if configMap.ResourceVersion != "" {
+		configMap.ResourceVersion = currentConfigMap.ResourceVersion
+	}
+	return rclient.Update(ctx, configMap)
+}
+
+func reconcileSVC(ctx context.Context, rclient client.Client, crdName string, svc *corev1.Service) error {
```
Kirill-Garbar, line 71: When I use camelCase names, usually I use them like this: reconcileSvc. Although I used it before the same: reconcileSVC. Why I changed my mind.
I google best practises and found one explanation.

Camel case and snake case should be able to be converted automatically to each other.

reconcileSVC will be converted to reconcile_s_v_c, because every capital letter in camelCase starts new word. And svc is one word. So in my opinion even abbreviations should not be written in capital letters.

This is not a blocking issue, just a comment to a code style, although we do not have common guide.

Kirill-Garbar (approved):

internal/controller/etcdcluster_controller.go
```diff
@@ -132,177 +131,24 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 func (r *EtcdClusterReconciler) ensureClusterObjects(
 	ctx context.Context, cluster *etcdaenixiov1alpha1.EtcdCluster, isClusterInitialized bool) error {
 	// 1. create or update configmap <name>-cluster-state
-	if err := r.ensureClusterStateConfigMap(ctx, cluster, isClusterInitialized); err != nil {
+	if err := factory.CreateOrUpdateClusterStateConfigMap(ctx, cluster, isClusterInitialized, r.Client, r.Scheme); err != nil {
```
aobort, line 134: Not a blocker, just slight concerns about the naming: despite of the argument name "isClusterInitialized", the real value which is passed to function reflects the rediness state of the cluster (https://github.com/aenix-io/etcd-operator/blob/feat/refactor-resource-generation/internal/controller/etcdcluster_controller.go#L87-L100)

```go
	isClusterReady := false
	sts := &appsv1.StatefulSet{}
	err = r.Get(ctx, client.ObjectKey{
		Namespace: instance.Namespace,
		Name:      instance.Name,
	}, sts)
	if err == nil {
		isClusterReady = sts.Status.ReadyReplicas == *sts.Spec.Replicas
	}

	if err := r.ensureClusterObjects(ctx, instance, isClusterReady); err != nil {
		logger.Error(err, "cannot create Cluster auxiliary objects")
		return r.updateStatusOnErr(ctx, instance, fmt.Errorf("cannot create Cluster auxiliary objects: %w", err))
	}
```
Which might lead to misunderstanding and confusion

internal/controller/factory/builders.go
```diff
@@ -47,3 +48,42 @@ func reconcileSTS(ctx context.Context, rclient client.Client, crdName string, st
...
+		if errors.IsNotFound(err) {
+			logger.V(2).Info("creating new configMap", "cm_name", configMap.Name, "crd_object", crdName)
+			return rclient.Create(ctx, configMap)
+		}
+		return fmt.Errorf("cannot get existing configMap: %s, for crd_object: %s, err: %w", configMap.Name, crdName, err)
+	}
+	configMap.Annotations = labels.Merge(currentConfigMap.Annotations, configMap.Annotations)
+	if configMap.ResourceVersion != "" {
+		configMap.ResourceVersion = currentConfigMap.ResourceVersion
+	}
```
aobort, line 67: According to field description:
**Populated by the system. Read-only.**

What's the goal of such assignment? The same for all auxiliary resources: services, sts.

sergeyshevch: We always apply only newly generated resources. So for update call we need to reflect this value to new object from current one

aobort: @sergeyshevch I got your point. However:
- the `configMap` is constructed on, lets say, previous step without `ResourceVersion`. Hence the check whether it is empty is redundant
- you're absolutely right that if we'll try to update outdated resource version it will result in an error. But it seems, that if we'll call update on the object constructed from scratch, then update will be applied to the latest version (in case the object with defined namespace/name is already exist) automagically. I've tested it on k8s v1.29.0, works like a charm without setting `ResourceVersion` manually. But I'm not sure about previous versions.

In total: it's not a blocker, if you consider it's ok to proceed without any changes - feel free to resolve the conversation. Just shared some thoughts.

sergeyshevch: Good point. This code initially was copied from VM Operator. In my understanding it shouldn't work without copy of `ResourceVersion`

I guess for not we will stay with this implementation. I will create issue to test and revisit it later

internal/controller/etcdcluster_controller.go
sergeyshevch: Updated naming

internal/controller/factory/builders.go
sergeyshevch: https://github.com/aenix-io/etcd-operator/issues/90 

FollowUp issue to research there more in-deep

Issue #79, opened 2024-03-23
Improve conditions processing


aobort:
- If conditions list is empty, fill it with all possible conditions in "False" state. This will allow to observe which phases cluster object should pass and already passed before it's in ready state (https://github.com/aenix-io/etcd-operator/blob/main/internal/controller/etcdcluster_controller.go#L75-L84).
- Move initiail conditions' list filling and further condition(s) state update to separate function. This will decrease complexity of `Reconcile` method, improve code readability and make it easier to add new conditions. (https://github.com/aenix-io/etcd-operator/blob/main/internal/controller/etcdcluster_controller.go#L102-L125)
- Update condition's `lastTransitionTime` only if the status changed

PR #92, opened 2024-03-28
improve EtcdCluster conditions processing

aobort:
Fixes: #79

### DIFF
--- a/internal/controller/factory/configMap.go
+++ b/internal/controller/factory/configMap.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/aenix-io/etcd-operator/internal/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -36,7 +37,6 @@ func GetClusterStateConfigMapName(cluster *etcdaenixiov1alpha1.EtcdCluster) stri
 func CreateOrUpdateClusterStateConfigMap(
 	ctx context.Context,
 	cluster *etcdaenixiov1alpha1.EtcdCluster,
-	isClusterReady bool,
 	rclient client.Client,
 	rscheme *runtime.Scheme,
 ) error {
@@ -63,7 +63,7 @@ func CreateOrUpdateClusterStateConfigMap(
 		},
 	}
 
-	if isClusterReady {
+	if util.EtcdClusterIsReady(cluster) {
 		// update cluster state to existing
 		configMap.Data["ETCD_INITIAL_CLUSTER_STATE"] = "existing"
 	}
--- a/internal/util/auxutil.go
+++ b/internal/util/auxutil.go
@@ -1,0 +1,88 @@
+package util
+
+import (
+	"slices"
+
+	etcdaenixiov1alpha1 "github.com/aenix-io/etcd-operator/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EtcdClusterIsReady returns true is condition "Ready" has status equals to "True", otherwise false.
+func EtcdClusterIsReady(cluster *etcdaenixiov1alpha1.EtcdCluster) bool {
+	idx := slices.IndexFunc(cluster.Status.Conditions, func(condition metav1.Condition) bool {
+		return condition.Type == etcdaenixiov1alpha1.EtcdConditionReady
+	})
+	if idx == -1 {
+		return false
+	}
+	return cluster.Status.Conditions[idx].Status == metav1.ConditionTrue
+}
+
+// FillConditions fills EtcdCluster .status.Conditions list with all available conditions in "False" status.
+func FillConditions(cluster *etcdaenixiov1alpha1.EtcdCluster) {
+	conditions := []string{
+		etcdaenixiov1alpha1.EtcdConditionInitialized,
+		etcdaenixiov1alpha1.EtcdConditionReady,
+	}
+
+	for _, c := range conditions {
+		SetCondition(cluster, c, false)
+	}
+}
+
+// SetCondition sets the condition of defined type to desired state.
+func SetCondition(
+	cluster *etcdaenixiov1alpha1.EtcdCluster,
+	conditionType string,
+	conditionStatus bool,
+) {
+	status := metav1.ConditionFalse
+	if conditionStatus {
+		status = metav1.ConditionTrue
+	}
+
+	idx := slices.IndexFunc(cluster.Status.Conditions, func(condition metav1.Condition) bool {
+		return condition.Type == conditionType
+	})
+
+	reason, message := conditionReasonMessage(conditionType, status)
+	condition := metav1.Condition{
+		Type:               conditionType,
+		Status:             status,
+		ObservedGeneration: cluster.Generation,
+		LastTransitionTime: metav1.Now(),
+		Reason:             reason,
+		Message:            message,
+	}
+
+	if idx == -1 {
+		cluster.Status.Conditions = append(cluster.Status.Conditions, condition)
+		return
+	}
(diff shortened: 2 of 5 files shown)


internal/util/auxutil.go
```diff
@@ -0,0 +1,88 @@
...
+
+	for _, c := range conditions {
+		SetCondition(cluster, c, false)
+	}
+}
+
+// SetCondition sets the condition of defined type to desired state.
+func SetCondition(
+	cluster *etcdaenixiov1alpha1.EtcdCluster,
+	conditionType string,
```
sircthulhu, line 36: I would use `etcdaenixiov1alpha1.EtcdCondType` here

internal/util/auxutil.go
```diff
@@ -0,0 +1,88 @@
...
+	case conditionType == etcdaenixiov1alpha1.EtcdConditionInitialized && status == metav1.ConditionTrue:
+		reason = string(etcdaenixiov1alpha1.EtcdCondTypeInitComplete)
+		message = string(etcdaenixiov1alpha1.EtcdInitCondPosMessage)
+	case conditionType == etcdaenixiov1alpha1.EtcdConditionReady && status == metav1.ConditionFalse:
+		reason = string(etcdaenixiov1alpha1.EtcdCondTypeStatefulSetNotReady)
+		message = string(etcdaenixiov1alpha1.EtcdReadyCondNegMessage)
+	case conditionType == etcdaenixiov1alpha1.EtcdConditionReady && status == metav1.ConditionTrue:
+		reason = string(etcdaenixiov1alpha1.EtcdCondTypeStatefulSetReady)
+		message = string(etcdaenixiov1alpha1.EtcdReadyCondPosMessage)
+	}
```
sircthulhu, line 86: This won't let add more options for negative statuses. For example after writing custom pod controller, reconciler can fail on creating pods because of resource quota or range limit. I think we will need to add custom message here. Let's pass everywhere message as just string without consts

internal/util/auxutil.go
sircthulhu, line 1: why `auxutil`? I would name it something like `conditions.go`

sircthulhu (changes requested):

internal/util/auxutil.go
aobort: I would move there auxiliary functions like those used in resource builders to construct names, 'cuz they might be helpful in future testing. However it's out of scope of this PR/issue it closes

internal/util/auxutil.go
aobort: It make sense, however at the moment it will be over engineering to take into account such cases, which are not even designed.
From my perspective, in the future it will be reasonable to make setting of condition as something like wrapper on the cluster's "phase" processing functions. This will allow us to make conditions' reasons & messages accurate and meaningful.
Also, I'd prefer not to put all possible failure reason-message into "Ready" state, but add conditions like "PodsScheduled" etc.

sircthulhu: It's not over engineering. It will simplify code:
- there won't be implicit selecting of messages for type
- it will be simpler to add new message for any type at any time

aobort: Hm. Ok then, wdyt about adding condition builder? Which will result in something like

```go
// etcdcluster_reconciler.go
...

util.setCondition(cluster, condition().
    WithType(...).
    WithStatus(...).
    WithReason(...).
    WithMessage(...)

...
```

sircthulhu: I like your suggestion :) 
It will be simpler as of helpers logic, it's just one parameter and it will be as powerful as 4 separate parameters

internal/util/auxutil.go
Kirill-Garbar: As I understand we are going to have new package **util**. I think we need to agree on the name and what is stored there. The place where different functions are stored I usually call common (not voting for the name) - it is just an option.

Also I read recommendations to avoid common names (common, utils, etc...), but I do not know how to avoid that if we have something really common. However I see in the package pretty concrete functions that are related to status conditions. I would not place them to utils at all. What's opinion of the others? Name the package differently. Or we can name it **utils** and then reorganize the code later when more functions exist in util package.

aobort: As I wrote in the comments thread below, I'd like to move to utils package all auxiliary functions, like objects' name constructors. I bet we'll have more such functions in the future and I'd prefer to keep them in one place.
What about naming - well, I call such packages as "util" in other projects, so it just a habit.
What about recommendations you've read: I believe it refers to exportable packages. Since the package is 'internal' I bet we can use any name considered to be convenient and meaningful for us.
From my perspective name 'util' perfectly reflects what the package contains.

internal/util/auxutil.go
```diff
@@ -0,0 +1,88 @@
+package util
+
+import (
+	"slices"
+
+	etcdaenixiov1alpha1 "github.com/aenix-io/etcd-operator/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EtcdClusterIsReady returns true is condition "Ready" has status equals to "True", otherwise false.
```
Kirill-Garbar, line 10: // EtcdClusterIsReady returns true **if** condition "Ready" has status **equal** to "True", otherwise false.

Is this correct comment?

internal/util/auxutil.go
```diff
@@ -0,0 +1,88 @@
...
+		cluster.Status.Conditions = append(cluster.Status.Conditions, condition)
+		return
+	}
+	if cluster.Status.Conditions[idx].Status == status {
+		return
+	}
+	cluster.Status.Conditions[idx] = condition
+}
+
+func conditionReasonMessage(conditionType string, status metav1.ConditionStatus) (string, string) {
```
Kirill-Garbar, line 68: Could you please add a comment to this function as well even if it is not exported?

aobort: This function will be removed after code will be updated following to @sircthulhu review.

internal/util/auxutil.go
aobort: SetCondition function updated as agreed in the comments thread below

internal/controller/factory/condition.go
```diff
@@ -0,0 +1,95 @@
...
+) {
+	idx := slices.IndexFunc(cluster.Status.Conditions, func(c metav1.Condition) bool {
+		return c.Type == condition.Type
+	})
+
+	if idx == -1 {
+		cluster.Status.Conditions = append(cluster.Status.Conditions, condition)
+		return
+	}
+	if cluster.Status.Conditions[idx].Status == condition.Status {
```
sircthulhu, line 91: If something changes in `Reason` without changing status, this change will not be reflected. It is not a common case but it will be hard to find a bug

internal/controller/factory/configMap.go
```diff
@@ -74,3 +74,14 @@ func CreateOrUpdateClusterStateConfigMap(
 
 	return reconcileConfigMap(ctx, rclient, cluster.Name, configMap)
 }
+
+// etcdClusterIsReady returns true if condition "Ready" has status equal to "True", otherwise false.
+func etcdClusterIsReady(cluster *etcdaenixiov1alpha1.EtcdCluster) bool {
```
sircthulhu, line 79: I would call it `isEtcdClusterReady`

sircthulhu (changes requested):

internal/controller/factory/condition.go
aobort: Make sense

internal/controller/factory/configMap.go
Kirill-Garbar: We have function clusterIsReady, that checks amount of ready replicas. And I would say that naming is not clear what is what :) That one function checks condition. Another function checks replicas.

I wild rename this function completely. Like CheckReadyCondition

aobort: Make sense

sircthulhu (approved):
LGTM

Kirill-Garbar (approved):

PR #93, opened 2024-03-28
Add PDB to EtcdCluster spec and implement validation webhhoks

### LINKED ISSUE #61: Make an option to create PDB for cluster
I think we should add an option to create PDB for cluster.
```diff
 ---
 apiVersion: etcd.aenix.io/v1alpha1
 kind: EtcdCluster
 metadata:
   name: test
   namespace: ns1
 spec:
   image: "quay.io/coreos/etcd:v3.5.12"
   replicas: 3
  storage:
    persistence: true # default: true, immutable
    storageClass: local-path
    size: 10Gi
+ enablePDB:  true
 status:
   conditions:
   - lastProbeTime: null
 	lastTransitionTime: "2024-03-06T18:39:45Z"
 	status: "True"
 	type: Ready
```

By default it should be `true` meaning that PDB resource will be created with `maxUnavailable` field equal to the maximum number of members can die without losing quorum

sircthulhu:
Added PDB to spec, implemented webhooks and tests.

Fixes #61


api/v1alpha1/etcdcluster_types.go
```diff
@@ -66,6 +70,11 @@ type EtcdCluster struct {
 	Status EtcdClusterStatus `json:"status,omitempty"`
 }
 
+// calculateQuorumSize returns minimum quorum size for current number of replicas
+func (r *EtcdCluster) calculateQuorumSize() int {
+	return int(math.Ceil(float64(*r.Spec.Replicas) / 2.))
```
Kirill-Garbar, line 75: What will be the quorum of etcd cluster of 3 and 4 nodes?
For 3 nodes it will be 2. For 4 nodes - 3.
As I understand current formula calculates 2 for 3 nodes cluster and 2 for 4 nodes cluster.

sircthulhu: Thank you for noticing! You're absolulely right. Fixed it :)

api/v1alpha1/etcdcluster_types.go
```diff
@@ -66,6 +70,16 @@ type EtcdCluster struct {
 	Status EtcdClusterStatus `json:"status,omitempty"`
 }
 
+// calculateQuorumSize returns minimum quorum size for current number of replicas
+func (r *EtcdCluster) calculateQuorumSize() int {
+	replicas := *r.Spec.Replicas
+	if replicas%2 == 0 {
```
Kirill-Garbar, line 76: I am not sure that you changed it correctly.
6 replicas
6-1=5
5/2=2.5
Ceil(2.5)=3

And it should be 4.

If you change "replicas - 1" to "replicas + 1" it will be correct I think.

sircthulhu: Thank you!

config/crd/bases/etcd.aenix.io_etcdclusters.yaml
```diff
@@ -39,6 +39,86 @@ spec:
           spec:
             description: EtcdClusterSpec defines the desired state of EtcdCluster
             properties:
+              podDisruptionBudget:
```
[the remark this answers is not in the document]
sircthulhu: Could you explain please?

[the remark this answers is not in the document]
Kirill-Garbar: I mean that sample CR should include PDB settings. It will be an example usage and anybody can install easily install it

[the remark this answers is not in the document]
sircthulhu: Basic PDB configuration is `spec.podDisruptionBudget: {}`, it will generate PDB resource with `minAvailable` configuration according to the number of replicas in the cluster. I don't think more options in sample is necessary. We can add them later to docs

api/v1alpha1/etcdcluster_types.go
```diff
@@ -192,6 +208,30 @@ type EmbeddedPersistentVolumeClaim struct {
...
 
+// EmbeddedPodDisruptionBudget describes PDB resource for etcd cluster members
+type EmbeddedPodDisruptionBudget struct {
+	// EmbeddedMetadata contains metadata relevant to an EmbeddedResource.
+	// +optional
+	EmbeddedObjectMetadata `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	// Spec defines the desired characteristics of a PDB.
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets
+	//+optional
+	Spec PodDisruptionBudgetSpec `json:"spec"`
```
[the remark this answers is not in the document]
sircthulhu: As we discussed internally, it's a bad idea to add fields that controller will ignore. It will confuse user trying to add `selector` for example

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -106,3 +121,66 @@ func (r *EtcdCluster) ValidateDelete() (admission.Warnings, error) {
 	etcdclusterlog.Info("validate delete", "name", r.Name)
 	return nil, nil
 }
+
+// validatePdb validates PDB fields
+func (r *EtcdCluster) validatePdb() (admission.Warnings, field.ErrorList) {
```
Kirill-Garbar, line 126: Internally discussed and agreed that pdb percentage validation and creation is missing. Waiting for the implementation.

config/crd/bases/etcd.aenix.io_etcdclusters.yaml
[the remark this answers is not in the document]
Kirill-Garbar: It is fine. This comment was written before sample CR was changed and pdb creation was implemented.

api/v1alpha1/etcdcluster_types.go
```diff
@@ -74,6 +78,11 @@ type EtcdCluster struct {
 	Status EtcdClusterStatus `json:"status,omitempty"`
 }
 
+// CalculateQuorumSize returns minimum quorum size for current number of replicas
+func (r *EtcdCluster) CalculateQuorumSize() int {
+	return int(*r.Spec.Replicas)/2 + 1
```
sergeyshevch, line 83: It can nil pointer error. For type safety I guess we need to check replicas for nil firstly

api/v1alpha1/etcdcluster_types.go
```diff
@@ -200,6 +209,30 @@ type EmbeddedPersistentVolumeClaim struct {
...
+	//+optional
+	Spec PodDisruptionBudgetSpec `json:"spec"`
+}
+
+type PodDisruptionBudgetSpec struct {
+	// MinAvailable describes minimum ready replicas. If both are empty, controller will implicitly
+	// calculate MaxUnavailable based on number of replicas
+	// Mutually exclusive with MaxUnavailable.
+	// +optional
+	MinAvailable *intstr.IntOrString `json:"minAvailable,omitempty"`
```
sergeyshevch, line 228: Can we do oneof validation with kubebuilder annotations here?

api/v1alpha1/etcdcluster_types_test.go
```diff
@@ -0,0 +1,22 @@
+package v1alpha1
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/utils/ptr"
+)
+
+var _ = Context("CalculateQuorumSize", func() {
+	It("should return correct result for odd number of replicas", func() {
```
sergeyshevch, line 10: It also good idea to test with nil value in replicas

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -106,3 +121,66 @@ func (r *EtcdCluster) ValidateDelete() (admission.Warnings, error) {
...
+	if r.Spec.PodDisruptionBudget == nil {
+		return nil, nil
+	}
+	pdb := r.Spec.PodDisruptionBudget
+	var warnings admission.Warnings
+	var allErrors field.ErrorList
+	if pdb.Spec.MinAvailable != nil && pdb.Spec.MaxUnavailable != nil {
```
sergeyshevch, line 133: Can we implement this with kubebuilder annotations? We don't need to support code if controller runtime already has capabilities to do it. Or maybe we can use CEL for such purpose

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -106,3 +121,66 @@ func (r *EtcdCluster) ValidateDelete() (admission.Warnings, error) {
...
+		allErrors = append(allErrors, field.Invalid(
+			field.NewPath("spec", "podDisruptionBudget", "minAvailable"),
+			pdb.Spec.MinAvailable.IntValue(),
+			"minAvailable is mutually exclusive with maxUnavailable"),
+		)
+	} else {
+		minQuorumSize := r.CalculateQuorumSize()
+		if pdb.Spec.MinAvailable != nil {
```
sergeyshevch, line 141: MinAvailable can contain value such as `50%`. Looks like it won't pass validaition here

api/v1alpha1/etcdcluster_webhook_test.go
```diff
@@ -81,8 +89,87 @@ var _ = Describe("EtcdCluster Webhook", func() {
 				},
 			}
 			w, err := etcdCluster.ValidateCreate()
-			gomega.Expect(err).To(gomega.Succeed())
-			gomega.Expect(w).To(gomega.BeEmpty())
+			Expect(err).To(Succeed())
+			Expect(w).To(BeEmpty())
+		})
+		It("Should admit enabled empty PDB", func() {
```
sergeyshevch, line 95: Looks like this testcase already too big. Can we test validatePdb method separately instead?

internal/controller/factory/pdb.go
```diff
@@ -0,0 +1,71 @@
...
+			MinAvailable:   cluster.Spec.PodDisruptionBudget.Spec.MinAvailable,
+			MaxUnavailable: cluster.Spec.PodDisruptionBudget.Spec.MaxUnavailable,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: NewLabelsBuilder().WithName().WithInstance(cluster.Name).WithManagedBy(),
+			},
+			UnhealthyPodEvictionPolicy: ptr.To(v1.IfHealthyBudget),
+		},
+	}
+	// if both are nil, calculate minAvailable based on number of replicas in the cluster
+	if pdb.Spec.MinAvailable == nil && pdb.Spec.MaxUnavailable == nil {
```
sergeyshevch, line 62: Looks like this is an impossible case. If none of min/max if specified. `cluster.Spec.PodDisruptionBudget` will be nil

sergeyshevch (changes requested):

api/v1alpha1/etcdcluster_types_test.go
sircthulhu: Nil value is impossible due to kubebuilder marker
```go
// +kubebuilder:default:=3
```

internal/controller/factory/pdb.go
sircthulhu: PodDisruptionBudget is pointer, so PDB will be created if `.spec.podDisruptionBudget: {}`

sergeyshevch: I guess we shouldn't create PDB in such case. It still empty

api/v1alpha1/etcdcluster_types_test.go
sergeyshevch: Then it's strange to have this value as a pointer in struct. If it always prefilled. But I guess it's out of scope of this PR. We can discuss it later

internal/controller/factory/pdb.go
Kirill-Garbar: In my opinion logic looks fine.
If we do not define section at all - do not create pdb. Define section empty - define with reasonable defaults. Define custom pdb - custom pdb is created.

api/v1alpha1/etcdcluster_webhook.go
sircthulhu: There's no kubebuilder marker for `one of`

kvaps:
@Kirill-Garbar could you please review this?

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -106,3 +124,97 @@ func (r *EtcdCluster) ValidateDelete() (admission.Warnings, error) {
...
+			pdb.Spec.MinAvailable.IntValue(),
+			"minAvailable is mutually exclusive with maxUnavailable"),
+		)
+		return nil, allErrors
+	}
+
+	minQuorumSize := r.CalculateQuorumSize()
+	if pdb.Spec.MinAvailable != nil {
+		minAvailable := pdb.Spec.MinAvailable.IntValue()
+		if pdb.Spec.MinAvailable.IntVal == 0 && pdb.Spec.MinAvailable.IntValue() == 0 {
```
Kirill-Garbar, line 149: On the previous line you defined and calculated minAvailable. Does it make sense to use the this variable in the condition?

Kirill-Garbar: Also.

As I understand we should check is min available a string and then go through percentage calculations. If min available is zero, then we will enter this condition validation will probably fail.

Kirill-Garbar (approved):

PR #94, opened 2024-03-28
fix: use status condition for setting cluster configmap state

sircthulhu:
Before that only current statefulset readiness status defined cluster state in configmap.

Checked on local Kind cluster: when using PVC, cluster members can successfully restart without errors. Cluster state in configmap is changed correctly.

### DIFF
--- a/internal/controller/etcdcluster_controller.go
+++ b/internal/controller/etcdcluster_controller.go
@@ -107,11 +107,11 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				Complete())
 		}
 	} else {
-		reason := etcdaenixiov1alpha1.EtcdCondTypeStatefulSetReady
-		message := etcdaenixiov1alpha1.EtcdReadyCondPosMessage
-		if !clusterReady {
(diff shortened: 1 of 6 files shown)


internal/controller/etcdcluster_controller.go
```diff
@@ -95,11 +95,31 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				WithMessage(string(etcdaenixiov1alpha1.EtcdReadyCondPosMessage)).
+				Complete())
+		}
+	} else {
+		reason := etcdaenixiov1alpha1.EtcdCondTypeStatefulSetReady
+		message := etcdaenixiov1alpha1.EtcdReadyCondPosMessage
+		if !clusterReady {
+			reason = etcdaenixiov1alpha1.EtcdCondTypeStatefulSetNotReady
+			message = etcdaenixiov1alpha1.EtcdReadyCondNegMessage
+		}
```
aobort, line 115: WDYT about swapping the positive and negative reason-message pairs to avoid condition inversion?

sircthulhu: It makes sense, thank you!

kvaps (approved):

aobort (approved):

PR #95, opened 2024-03-29
enh: Add common commands for local development

sircthulhu:
Added commands for local development.
- `make docker-load` builds and loads docker image to local kind cluster
- `make redeploy` rebuilds, loads, deploys manifests and forces controller pods to recreate

Also, I changed default IMG in Makefile to be `ghcr.io`, so that it won't be necessary to change it every time for local testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,16 +151,16 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
 # Build and upload docker image to the local Kind cluster
 .PHONY: docker-load
 docker-load: docker-build
-	kind load docker-image ${IMG}
+	kind load docker-image ${IMG} --name etcd-operator-kind
 
 # Redeploy controller with new docker image
 .PHONY: redeploy
 redeploy: manifests kustomize docker-build docker-load
+	$(KUBECTL) config use-context kind-etcd-operator-kind
 	# deploy configs
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
 	# force recreate pods
-	$(KUBECTL) delete pod -n etcd-operator-system --all
+	$(KUBECTL) rollout restart -n etcd-operator-system deploy/etcd-operator-controller-manager
 
 ##@ Dependencies
 
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,20 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
+# Build and upload docker image to the local Kind cluster
+.PHONY: docker-load
+docker-load: docker-build
+	kind load docker-image ${IMG}
+
+# Redeploy controller with new docker image
+.PHONY: redeploy
+redeploy: manifests kustomize docker-build docker-load
+	# deploy configs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -148,6 +148,20 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
...
+	kind load docker-image ${IMG}
+
+# Redeploy controller with new docker image
+.PHONY: redeploy
+redeploy: manifests kustomize docker-build docker-load
+	# deploy configs
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+	# force recreate pods
+	$(KUBECTL) delete pod -n etcd-operator-system --all
```
Kirill-Garbar, line 163: Why delete and not rollout restart? Somebody will test zerotime deployment in the future.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: ghcr.io/aenix-io/etcd-operator
```
Kirill-Garbar, line 7: There is a guide how to develop locally in the repo. I remember that there is a reference to image retagging (simple guide), but from the mobile phone I can't tell exactly what needs to be changed if needed.

Could you please look it through?

aobort, line 7: @sircthulhu the reason, why kustomization does not work for image is in "name" field: we need either change it to `ghcr.io/aenix-io/etcd-operator` (as it is in manager.yaml) in kustomization.yaml or change the image name to `controller` in deployment manifest (manager.yaml)

hiddenmarten (dismissed):

config/manager/kustomization.yaml
sircthulhu: There's a line in Makefile
```sh
cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
```
it creates an element in `config/manager/kustomization.yaml` `images` list that replcases `controller` image to provided `${IMG}`. If this element already exists, nothing will be changed. If not, new element will be prepended to the list

Makefile
```diff
@@ -148,6 +148,20 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
+# Build and upload docker image to the local Kind cluster
+.PHONY: docker-load
+docker-load: docker-build
+	kind load docker-image ${IMG} --name etcd-operator-kind
```
hiddenmarten, line 154: I believe it's unnecessary to use a fixed name like `etcd-operator-kind`. Instead, using the default name or allowing configuration through an environment variable could be more flexible. Note that the latter option would necessitate an extra setup step for kind creation.

sircthulhu: I used `name` from [HOW-TO-START-DEVELOPMENT.md](https://github.com/aenix-io/etcd-operator/blob/main/HOW-TO-START-DEVELOPMENT.md). Do you think I should pass it through a Makefile env?

hiddenmarten: Yep, I suppose it should be like that.

Honestly, I wanted to make documentation [about development process](https://github.com/aenix-io/etcd-operator/blob/main/HOW-TO-START-DEVELOPMENT.md) smaller, just put steps for creating `kind` cluster with all required installation as a sequence of Makefile runs. You can skip this comment, I hope I'll have time to do it separately.

hiddenmarten (approved):

kvaps (approved):

Issue #82, opened 2024-03-25
Add ability to customize probe settings in podSpec


sergeyshevch:
Follow up for: https://github.com/aenix-io/etcd-operator/pull/72#discussion_r1536817322

We need to add new field to podSpec, agree with maintainers on further spec. And implement it as DoD for this issue

PR #97, opened 2024-03-29
Feature/customize probe settings in pod spec

ilya-hontarau:
Issue: #82


internal/controller/factory/statefulset.go
```diff
@@ -221,3 +194,65 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+
+	if probe.InitialDelaySeconds != 0 {
+		defaultProbe.InitialDelaySeconds = probe.InitialDelaySeconds
+	}
+
+	if probe.PeriodSeconds != 0 {
+		defaultProbe.PeriodSeconds = probe.PeriodSeconds
+	}
+
+	if probe.HTTPGet != nil {
```
sircthulhu, line 253: Probes can be of other types and they're mutually exclusive. I think here should be a check for any of them in `probe`

internal/controller/factory/statefulset.go
```diff
@@ -221,3 +194,65 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+				Port: intstr.FromInt32(2379),
+			},
+		},
+		InitialDelaySeconds: 5,
+		PeriodSeconds:       5,
+	}
+	return mergeWithDefaultProbe(probe, defaultProbe)
+}
+
+func mergeWithDefaultProbe(probe *corev1.Probe, defaultProbe *corev1.Probe) *corev1.Probe {
```
sircthulhu, line 240: I think usage of pointers is bad here as `defaultProbe` is altered and then returned. Function is not pure

internal/controller/factory/statefulset_test.go
```diff
@@ -129,6 +130,135 @@ var _ = Describe("CreateOrUpdateStatefulSet handler", func() {
...
+			}
+
+			sts := &appsv1.StatefulSet{}
+			err := CreateOrUpdateStatefulSet(ctx, etcdcluster, k8sClient, k8sClient.Scheme())
+			Expect(err).NotTo(HaveOccurred())
+
+			err = k8sClient.Get(ctx, typeNamespacedName, sts)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Checking the updated startup probe")
```
sircthulhu, line 214: Please use `By` as follows:
```go
By("Checking the updated startup probe", func () {
// all code goes here
})
```

api/v1alpha1/etcdcluster_types.go
```diff
@@ -164,6 +164,18 @@ type PodSpec struct {
 	// ExtraEnv are the extra environment variables to pass to the etcd container.
 	// +optional
 	ExtraEnv []corev1.EnvVar `json:"extraEnv,omitempty"`
+
+	// LivenessProbe defines liveness probe check for the pod.
```
sircthulhu, line 168: WDYT about adding something like "if empty, default probe will be used"?

sircthulhu (changes requested):

internal/controller/factory/statefulset.go
sircthulhu: I think it is rather good provided that user can customize image and add custom scripts for health checks

api/v1alpha1/etcdcluster_types.go
ilya-hontarau: It makes sense, the only question how do we want to leave here a reference to this default probe in the comment?

internal/controller/factory/statefulset.go
ilya-hontarau: Do you think we also need to validate if there is more than one probe?
> Probes can be of other types and they're mutually exclusive. I think here should be a check for any of them in probe

internal/controller/factory/statefulset_test.go
ilya-hontarau: Fixed my changes, but I see in other places the same pattern, maybe fix it in other pr

internal/controller/factory/statefulset.go
ilya-hontarau: Good notice, fixed

api/v1alpha1/etcdcluster_types.go
sircthulhu: I think we can mention endpoint we use for specific probe. For example, "By default /livez endpoint will be used" for liveness probe

@Kirill-Garbar do you have any ideas about that?

internal/controller/factory/statefulset.go
sircthulhu: I think yes, it should be "one of"

api/v1alpha1/etcdcluster_types.go
sergeyshevch: I like idea to use default liveness/readiness probes for etcd

internal/controller/factory/statefulset.go
```diff
@@ -221,3 +194,69 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+
+func getStartupProbe(probe *corev1.Probe) *corev1.Probe {
+	defaultProbe := corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: "/readyz?serializable=false",
+				Port: intstr.FromInt32(2379),
+			},
+		},
+		InitialDelaySeconds: 1,
```
sergeyshevch, line 206: I guess initial delay for startup probe can be 0

internal/controller/factory/statefulset.go
```diff
@@ -221,3 +194,69 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+
+func getReadinessProbe(probe *corev1.Probe) *corev1.Probe {
+	defaultProbe := corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: "/readyz",
+				Port: intstr.FromInt32(2379),
+			},
+		},
+		InitialDelaySeconds: 5,
```
sergeyshevch, line 220: We don't need initial delay here if we have startup probe

internal/controller/factory/statefulset.go
```diff
@@ -221,3 +194,69 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+
+func getLivenessProbe(probe *corev1.Probe) *corev1.Probe {
+	defaultProbe := corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: "/livez",
+				Port: intstr.FromInt32(2379),
+			},
+		},
+		InitialDelaySeconds: 5,
```
sergeyshevch, line 234: We don't need initial delay here if we have startup probe

internal/controller/factory/statefulset_test.go
```diff
@@ -129,8 +130,144 @@ var _ = Describe("CreateOrUpdateStatefulSet handler", func() {
...
+					},
+					InitialDelaySeconds: 5,
+					TimeoutSeconds:      1,
+					PeriodSeconds:       5,
+					SuccessThreshold:    1,
+					FailureThreshold:    3,
+				}))
+			})
+
+			By("Deleting the statefulset", func() {
```
sergeyshevch, line 184: Let's move this By to AfterEach. It already used in all cases

sergeyshevch (changes requested):

sergeyshevch:
Closes: #82

internal/controller/factory/statefulset_test.go
ilya-hontarau: Actually it isnt used in the case:
> "should fail on creating the statefulset with invalid owner reference"

internal/controller/factory/statefulset.go
ilya-hontarau: But we always have startup probe, right?

internal/controller/factory/statefulset.go
ilya-hontarau: But we always have startup probe, right?

internal/controller/factory/statefulset.go
ilya-hontarau: > I think yes, it should be "one of"

@sircthulhu  But k8s resources doesnt validate it, and other operators too, not sure do we want to do this?

sircthulhu: if kubernetes does not validate it, we can leave it as is

sircthulhu (approved):
lgtm

internal/controller/factory/statefulset.go
sergeyshevch: Yes I guess

internal/controller/factory/statefulset.go
sergeyshevch: Yes I guess

sergeyshevch (changes requested):

internal/controller/factory/statefulset.go
ilya-hontarau: okay, then i can remove this delay

internal/controller/factory/statefulset.go
ilya-hontarau: okay, then i can remove this delay

sergeyshevch (approved):

Issue #77, opened 2024-03-23
Validate extraArgs field in kind:EtcdCluster


sergeyshevch:
We should validate extraArgs field and prevent usage of args generated by operator itself.
Follow up for #72 #69

sircthulhu:
Should we check that etcd supports such flags or just reject collisions for flags used by operator?

sergeyshevch:
The initial request was just to reject collisions. I guess it will be enough for now

andreykont:
Which of 3 behavior  we want:
1. Validation web hook
2. status error
3. skip the unwanted argument

sergeyshevch:
I guess validation webhook error

andreykont:
[](https://github.com/aenix-io/etcd-operator/pull/99)

PR #99, opened 2024-03-31
Feature/77 validate extraargs

andreykont:
Closes: #77


internal/controller/factory/statefulset.go
```diff
@@ -214,10 +222,43 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+func ValidatePodExtraArgs(cluster *etcdaenixiov1alpha1.EtcdCluster) error {
+	if len(cluster.Spec.PodSpec.ExtraArgs) == 0 {
+		return nil
+	}
+
+	baseArgs := argsFromSliceToMap(generateBaseEtcdArgs(cluster))
+
+	for name := range cluster.Spec.PodSpec.ExtraArgs {
+		if strings.HasPrefix(name, "-") {
+			return fmt.Errorf("the extra argument shoudn't have dashes, flag: %s", name)
```
aobort, line 253: Extra arg should not start from dash and have trailing dash, but it can contain dashes in the middle.

internal/controller/factory/statefulset.go
```diff
@@ -214,10 +222,43 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+		}
+
+		flag := "--" + name
+
+		if _, exists := baseArgs[flag]; exists {
+			return fmt.Errorf("can't use base exta argument '%s' in .Spec.PodSpec.ExtraArgs", flag)
```
aobort, line 259: Typo: exta -> extra

internal/controller/factory/utils.go
```diff
@@ -0,0 +1,24 @@
...
+
+		if len(parts) > 1 {
+			value = parts[1]
+		}
+
+		m[flag] = value
+	}
+
+	return m
+}
```
aobort, line 24: I'd keep such specific auxiliary function as close as possible to the place where it is used: in the `statefulset.go`.

internal/webhooks/etcdcluster_webhook_test.go
```diff
@@ -14,24 +14,28 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v1alpha1
+package webhooks
 
 import (
 	. "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
+	"golang.org/x/net/context"
```
aobort, line 22: Why not just "context" package?

internal/webhooks/etcdcluster_webhook.go
```diff
@@ -0,0 +1,147 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/aenix-io/etcd-operator/api/v1alpha1"
+	"github.com/aenix-io/etcd-operator/internal/controller/factory"
+)
+
+// log is for logging in this package.
+var etcdclusterlog = logf.Log.WithName("etcdcluster-resource")
+
+type EtcdCluster struct{}
```
sircthulhu, line 41: What is the purpose of moving webhooks to `internal` package? I suppose webhooks depend on CRD structure and should stay in `v1alpha1` for now

aobort: @sircthulhu I had the same question. As far as I understand, webhook imports both packages: api and factory. Hence, since factory imports api, in case webhooks would be a part of api package there woulb be circular dependencies.
Which definitely can be solved be re-organising of new feature.
I'm with you on the point to keep initial structure as is.

internal/controller/factory/statefulset.go
andreykont: fixed.

internal/controller/factory/statefulset.go
andreykont: fixed

internal/controller/factory/utils.go
andreykont: function moved

internal/webhooks/etcdcluster_webhook.go
andreykont: Yep, to avoid circular imports

andreykont: Also this approach allows add some interfaces into 
```type EtcdCluster struct{} ```
during initialization.

File: internal/webhooks/etcdcluster_webhook_test.go (hunk 4 above)
Comment (andreykont) [reply internal/webhooks/etcdcluster_webhook_test.go:22]: It's my fault. Fixed

File: internal/webhooks/etcdcluster_webhook.go (hunk 5 above)
Comment (andreykont) [reply internal/webhooks/etcdcluster_webhook.go:41]: I can move webhooks files back. Then make isolation based on interface.
It will look like this:
```
// +k8s:deepcopy-gen=false
type FactoryValidator interface {
	ValidatePodExtraArgs(*EtcdCluster) error
}

// +k8s:deepcopy-gen=false
type EtcdClusterHook struct {
	FactoryValidator
}

// SetupWebhookWithManager will setup the manager to manage the webhooks
func (r *EtcdClusterHook) SetupWebhookWithManager(mgr ctrl.Manager) error {
	return ctrl.NewWebhookManagedBy(mgr).
		For(&EtcdCluster{}).
		WithValidator(r).
		WithDefaulter(r).
		Complete()
}
```
Then  I need to make validation function  as a method.
```
type Validator struct{}

func (v Validator) ValidatePodExtraArgs(cluster *etcdaenixiov1alpha1.EtcdCluster) error
```

But, I don't like how it looks like...

Comment (andreykont) [reply internal/webhooks/etcdcluster_webhook.go:41]: Please help me to take a choice.

Comment (sircthulhu) [reply internal/webhooks/etcdcluster_webhook.go:41]: I would put `ValidatePodExtraArgs` to `etcdcluster_webhook.go` as it just performs validation, not logic for generating statefulset

Comment (andreykont) [reply internal/webhooks/etcdcluster_webhook.go:41]: The knowledge about how to build arguments belongs to business logic.
I want to store base arguments only in one packet to avoid business logic splitting. (DRY)
It seems that validation process belongs to the factory packet too.

There are three ways:
1. Repeat two absolutely the same sets of arguments in different packages. 
2. Move webhooks into particular package and use custom hooks  (as I did)
3. Use custom hooks isolated on interfaces. It needs to transform factory package to sets of methods.

Comment (sircthulhu) [reply internal/webhooks/etcdcluster_webhook.go:41]: > The knowledge about how to build arguments belongs to business logic.

I agree. But validation rules should belong to webhooks in CRD's package.

In `v1alpha1` we validate keys: reject prefix dashes and reject intersection with flags used by the operator. In `factory` we use keys and values from map to build flags for statefulset.

Can't find any logic duplications. Am I missing something?

Comment (sergeyshevch) [reply internal/webhooks/etcdcluster_webhook.go:41]: I like Idea to separate webhooks from API. Then we can setup API as separate go module that contains only definitions. If later someone want to include this module, he will depend only on API dependencies. Not all operator dependencies

**Review by sergeyshevch — CHANGES_REQUESTED** [review]

File: internal/webhooks/etcdcluster_webhook.go (hunk 5 above)
Comment (andreykont) [reply internal/webhooks/etcdcluster_webhook.go:41]: > In `v1alpha1` we validate keys: reject prefix dashes and reject intersection with flags used by the operator. In `factory` we use keys and values from map to build flags for statefulset.

That case implies store the flags  in two places.

**Review by sergeyshevch — APPROVED** [review]

File: internal/webhooks/etcdcluster_webhook.go (hunk 5 above)
Comment (sircthulhu) [reply internal/webhooks/etcdcluster_webhook.go:41]: @sergeyshevch I don't like moving webhooks to internal as for each api version there will be different webhooks. If we use interfaces and additional go module in api, users will pull whole project with it's dependencies. Do we really need this validation for extra args at all?

Operator must override flags and ignore if conflict happens


@andreykont 
> That case implies store the flags in two places.

Actually, we do not store this flags anywhere, we just use them in one place and say they're prohibited to use in another place. You can't do it any other way as validation is separate from usage even in one package. Checking validation after implementing changes to business logic is really common I think and it won't lead to problems. If it will, it's user's fault :)

**Review by sergeyshevch — APPROVED** [review]

**Comment by Kirill-Garbar** [comment]
It was discussed during 4th community meeting (02.04.24) and decided to leave webhooks in api directory.

**Main reason**: it is not a common pattern to place webhooks to internal directory. Operators that do the same could not be found.

See [minutes](https://docs.google.com/document/d/1coiWlHOVfPAIyKqtfpYLhq9eeRPHCuuMxmNmVfo9-sA/edit)

FYI @andreykont @sircthulhu @sergeyshevch @aobort @kvaps

File: internal/webhooks/etcdcluster_webhook.go (hunk 5 above)
Comment (Kirill-Garbar) [reply internal/webhooks/etcdcluster_webhook.go:41]: https://github.com/aenix-io/etcd-operator/pull/99#issuecomment-2032241368

File: api/v1alpha1/etcdcluster_webhook.go [hunk 6]
```diff
@@ -98,11 +102,42 @@ func (r *EtcdCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, er
...
+		"initial-advertise-peer-urls": {},
+		"data-dir":                    {},
+		"auto-tls":                    {},
+		"peer-auto-tls":               {},
+		"advertise-client-urls":       {},
+	}
+
+	for name := range cluster.Spec.PodSpec.ExtraArgs {
+		if strings.HasPrefix(name, "-") || strings.HasSuffix(name, "-") {
+			return fmt.Errorf("Extra arg should not start from dash and have trailing dash, "+
```
sircthulhu, line 132: should not start **with** dash

internal/controller/factory/statefulset.go
```diff
@@ -216,8 +222,15 @@ func generateEtcdCommand(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
 	for name, value := range cluster.Spec.PodSpec.ExtraArgs {
-		command = append(command, fmt.Sprintf("--%s=%s", name, value))
+		flag := "--" + name
+		if len(value) == 0 {
+			args = append(args, flag)
+
+			continue
+		}
+
+		args = append(args, fmt.Sprintf("%s=%s", flag, value))
```
sircthulhu, line 232: Do we need to sort keys here to be idempotent?

api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -98,11 +102,42 @@ func (r *EtcdCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, er
 		return warnings, err
 	}
 
-	return warnings, nil
+	return warnings, validateExtraArgs(r)
```
sircthulhu, line 105: I think we should use `allErrors` for possible errors returned from `validateExtraArgs` to return all validation errors at once

internal/controller/factory/statefulset.go
andreykont: There is tight dependency on sort order of arguments. I've got it when I tested it.

api/v1alpha1/etcdcluster_webhook.go
andreykont: Fixed. Now error looks like aggregate of errors:
Error from server (Forbidden): error when creating "config/samples/etcd.aenix.io_v1alpha1_etcdcluster.yaml": admission webhook "vetcdcluster.kb.io" denied the request: [Extra arg should not start with dash and have trailing dash, but it can contain dashes in the middle. flag: --xxx, can't use extra argument 'auto-tls' in .Spec.PodSpec.ExtraArgs because it generated by operator for cluster setup, can't use extra argument 'name' in .Spec.PodSpec.ExtraArgs because it generated by operator for cluster setup]

sircthulhu (approved):
lgtm

api/v1alpha1/etcdcluster_webhook.go
andreykont: Error on update:
# etcdclusters.etcd.aenix.io "etcdcluster-sample" was not valid:
# * spec.podSpec.extraArgs: Invalid value: map[string]string{"--xxx":"sdfdsf", "auto-tls":"", "name":"avkontya"}: [can't use extra argument 'auto-tls' in .Spec.PodSpec.ExtraArgs because it generated by operator for cluster setup, can't use extra argument 'name' in .Spec.PodSpec.ExtraArgs because it generated by operator for cluster setup, Extra arg should not start with dash and have trailing dash, but it can contain dashes in the middle. flag: --xxx]

Issue #90, opened 2024-03-27
Revisit ResourceVersion usage in resource factories


sergeyshevch:
@aobort in https://github.com/aenix-io/etcd-operator/pull/84#discussion_r1540079094 found that resource update correctly works without copying of `ResourceVersion` to newly generated structure in k8s 1.29+

We need to verify it for different k8s versions and remove if it isn't required

aobort:
How many versions prior the last one do we want to support and need to be tested?

sircthulhu:
@aobort according to [kubernetes releases](https://kubernetes.io/releases/), version 1.27 is going to be the end of line in June, 2024. I suppose we need to support all "alive" versions

aobort:
Removed copying of `ResourceVersion` to newly generated structure and run tests for k8s versions 1.27.1, 1.28.0, 1.29.0:

```shell
❯ for ver in {1.27.1,1.28.0,1.29.0}; do ENVTEST_K8S_VERSION=$ver make test; done
/Users/A117675684/Repos/aenix.io/etcd-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/A117675684/Repos/aenix.io/etcd-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
KUBEBUILDER_ASSETS="/Users/A117675684/Repos/aenix.io/etcd-operator/bin/k8s/1.27.1-darwin-arm64" go test $(go list ./... | grep -v /e2e) -coverprofile cover.out
?       github.com/aenix-io/etcd-operator/cmd   [no test files]
?       github.com/aenix-io/etcd-operator/test/utils    [no test files]
ok      github.com/aenix-io/etcd-operator/api/v1alpha1  4.283s  coverage: 14.0% of statements
ok      github.com/aenix-io/etcd-operator/internal/controller   6.520s  coverage: 50.0% of statements
ok      github.com/aenix-io/etcd-operator/internal/controller/factory   4.564s  coverage: 92.2% of statements
/Users/A117675684/Repos/aenix.io/etcd-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/A117675684/Repos/aenix.io/etcd-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
[...]

PR #100, opened 2024-04-01
add k8s versions matrix to test job

aobort:
Fixes: #90

### DIFF
--- a/.github/workflows/make-test.yaml
+++ b/.github/workflows/make-test.yaml
@@ -11,10 +11,13 @@ jobs:
   build:
     # Pull request has label 'ok-to-test' or the author is a member of the organization
     if: contains(github.event.pull_request.labels.*.name, 'ok-to-test') || contains(fromJSON('["COLLABORATOR", "MEMBER", "OWNER"]'), github.event.pull_request.author_association)
+    strategy:
+      matrix:
+        k8s_version: [1.27.1,1.28.0,1.29.0]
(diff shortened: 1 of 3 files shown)


.github/workflows/make-test.yaml
```diff
@@ -11,10 +11,13 @@ jobs:
   build:
     # Pull request has label 'ok-to-test' or the author is a member of the organization
     if: contains(github.event.pull_request.labels.*.name, 'ok-to-test') || contains(fromJSON('["COLLABORATOR", "MEMBER", "OWNER"]'), github.event.pull_request.author_association)
+    strategy:
+      matrix:
+        k8s_version: [1.27.1,1.28.0,1.29.0]
```
sergeyshevch, line 16: Maybe better to pickup last patch version?

sircthulhu (approved):
I would use latest patch version but i'm ok with that.
LGTM

.github/workflows/make-test.yaml
aobort: Well, I thought about it. My decision is due to the fact that the patch versions should be backwards compatible, hence it really does not matter which patch version is used in testing. Along with this, not all versions are available here https://storage.googleapis.com/kubebuilder-tools/ (where setup-envtest fetches assets).
Considering these two points, I decided to use the first patch version available, to not to track availability of the latest one.

Issue #109, opened 2024-04-02
[epic] API Design for EtcdCluster resource


kvaps:
We are going to release MVP (v0.1.0) and we need a stable spec. Here is the corrected version of the spec from this proposal: https://github.com/aenix-io/etcd-operator/issues/62 (original author: @sergeyshevch).

I'm going to use this meta-issue to link all parts for implementing this:

- [x] https://github.com/aenix-io/etcd-operator/pull/93
- [x] https://github.com/aenix-io/etcd-operator/pull/114
- [x] https://github.com/aenix-io/etcd-operator/issues/110
- [x] https://github.com/aenix-io/etcd-operator/issues/115
- [x] https://github.com/aenix-io/etcd-operator/issues/107
- [x] https://github.com/aenix-io/etcd-operator/issues/113

```yaml
---
apiVersion: etcd.aenix.io/v1alpha1
kind: EtcdCluster
metadata:
  name: test
  namespace: default
spec:
  replicas: 3
  options: # map[string]string
    election-timeout: "1000"
    max-wals: "5"
    max-snapshots: "5"

  storage:
    emptyDir: {} # core.v1.EmptyDirVolumeSource Ready k8s type
    volumeClaimTemplate:
      metadata:
        labels:
          env: prod
        annotations:
          example.com/annotation: "true"
      spec: # core.v1.PersistentVolumeClaimSpec Ready k8s type
        storageClassName: gp3
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 10Gi

  podTemplate:
    metadata:
      labels:
        env: prod
      annotations:
        example.com/annotation: "true"
    spec:
      imagePullSecrets:  # core.v1.LocalObjectReference Ready k8s type
      - name: myregistrykey
      serviceAccountName: default
      affinity: {} # core.v1.Affinity Ready k8s type
      nodeSelector: {} # map[string]string
      tolerations: [] # core.v1.Toleration Ready k8s type
      securityContext: {} # core.v1.PodSecurityContext Ready k8s type
      priorityClassName: "low"
      topologySpreadConstraints: [] # core.v1.TopologySpreadConstraint Ready k8s type
      terminationGracePeriodSeconds: 30 # int64
      schedulerName: "default-scheduler"
      runtimeClassName: "legacy"
      readinessGates: [] # core.v1.PodReadinessGate Ready k8s type
      containers: # []v1.Container
      - name: etcd
[...]

PR #114, opened 2024-04-02
Add EtcdCluster resource examples

kvaps:
following up https://github.com/aenix-io/etcd-operator/issues/109#issuecomment-2032847557

Here are examples for:

- Simple cluster with few replicas
- Cluster with custom image
- Cluster with persistent storage
- Cluster with resources
- Cluster with all field specified

### DIFF
--- a/examples/manifests/etcdcluster-full.yaml
+++ b/examples/manifests/etcdcluster-full.yaml
@@ -1,0 +1,78 @@
+---
+apiVersion: etcd.aenix.io/v1alpha1
+kind: EtcdCluster
+metadata:
+  name: test
+  namespace: default
+spec:
+  replicas: 3
+  options:
+    election-timeout: "1000"
+    max-wals: "5"
+    max-snapshots: "5"
+
+  storage:
+    emptyDir: {}
+    volumeClaimTemplate:
+      metadata:
+        labels:
+          env: prod
+        annotations:
+          example.com/annotation: "true"
+      spec:
+        storageClassName: gp3
+        accessModes: [ "ReadWriteOnce" ]
+        resources:
+          requests:
+            storage: 10Gi
+
+  podTemplate:
+    metadata:
+      labels:
+        env: prod
+      annotations:
+        example.com/annotation: "true"
+    spec:
+      imagePullSecrets:
+      - name: myregistrykey
+      serviceAccountName: default
+      affinity: {}
+      nodeSelector: {}
+      tolerations: []
+      securityContext: {}
+      priorityClassName: "low"
+      topologySpreadConstraints: []
+      terminationGracePeriodSeconds: 30
+      schedulerName: "default-scheduler"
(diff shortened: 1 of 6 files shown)


kvaps:
I’ll personally raise the question again whether we should replace:

`spec.pod` --> `spec.template`
`spec.service` --> `spec.serviceTemplate`
`spec.podDisruptionBudget` --> `spec.podDisruptionBudgetTemplate`

piraeus-operator and elk-operator use it this way

sircthulhu (approved):
lgtm

I would personally add `etcdcluster-with-default-pdb.yaml`
```yaml
---
apiVersion: etcd.aenix.io/v1alpha1
kind: EtcdCluster
metadata:
  name: test
  namespace: default
spec:
  replicas: 3
  podDisruptionBudget: {}
```

kvaps:
PR rebased after [internal discussion](https://t.me/etcd_operator/5601) to use Template suffixes for configuring Kubernetes entities.

Another advantage of using the Template suffixes is that it serves as a contract to adhere to the Kubernetes specifications.

What we have in storage and options is validated and processed by our logic, and the specifications there are strictly our own. However, what is in *Template follows the Kubernetes specifications and is intuitively understandable.

sergeyshevch (approved):

examples/manifests/etcdcluster-full.yaml
lllamnyp, line 1: In
```yaml
storage:
  emptyDir: {}
  volumeClaimTemplate: # ...
```
how do we know if an emptydir or a PVC will be used?

hiddenmarten (approved):

examples/manifests/etcdcluster-full.yaml
sircthulhu: Currently storage spec is the following:
```go
type StorageSpec struct {
	// EmptyDirVolumeSource to be used by the StatefulSets. If specified, used in place of any volumeClaimTemplate. More
	// info: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
	// +optional
	EmptyDir *corev1.EmptyDirVolumeSource `json:"emptyDir,omitempty"`
	// A PVC spec to be used by the StatefulSets.
	// +optional
	VolumeClaimTemplate EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate,omitempty"`
}
```

if emptyDir is null, then PVC. If emptyDir is not null, then emptyDir

sircthulhu (approved):

examples/manifests/etcdcluster-full.yaml
lllamnyp: So the manifest in the example is not really valid, as both `emptyDir` and `volumeClaimTemplate` are not null (emptyDir is an empty non-nil map)?

kvaps: yep, it should be OneOf

Issue #119, opened 2024-04-03
CRD updating in helm-chart using pre-commit


hiddenmarten:
For an easier development process it could be implemented in Makefile and added to pre-commit.

As a start point, we could use this script:

```bash
mkdir -p tmp && cd config/manager &&  \
kustomize edit set image ghcr.io/aenix-io/etcd-operator=ghcr.io/aenix-io/etcd-operator:latest && \
cd $OLDPWD && kustomize build config/default > tmp/manifest.yaml && \
cd tmp && yq -s '.kind + .metadata.name' --no-doc manifest.yaml
```

Issue #76, opened 2024-03-23
[epic] Secure communication with x509 certificates 


kvaps:
Secure communication between etcd members and etcd clients with x509 certificates.

This is a last task remained before we start using etcd-operator in our [cozystack](https://cozystack.io) platform.

- [ ] https://github.com/aenix-io/etcd-operator/pull/125

kvaps:
@gecube could you please elaborate on this, and write user stories and user spec for certificates

kvaps:
link https://github.com/aenix-io/etcd-operator/pull/87/files

Kirill-Garbar:
For now it was decided to create secret reference sections in the etcdCluster spec and expect secrets with certificates from the user of etcd-operator.

Basic spec from @lllamnyp 
```
spec:
  security:
    serverTLSSecretRef: # secretRef
      name: server-tls-secret
    clientCertAuth: true # bool
    trustedCAFile: # secretRef
      name: trusted-tls-secret
    peerTLSSecretRef: # secretRef
      name: peer-tls-secret
    peerClientCertAuth: true # bool
```

FYI @kvaps

Kirill-Garbar:
### Peer.ca
| Option | Description |
| ------ | ----------- |
| secretName  | Secret name of user-provided secret. If not specified then operator generates certificate by the spec below |
| metadata    | Metadata of generated secret. |
| duration    | Expiration time of generated secret. |
| renewBefore | Time period before expiration time when certificate will be reissued. |
| privateKey  | Private key configuration: algorithm and key size. |

### Peer.cert
| Option | Description |
| ------ | ----------- |
| secretName  | Secret name of user-provided secret. If not specified then operator generates certificate by the spec below. If peer.ca.secretName is provided, then this certificate is generated from the CA that was provided by the user. You can't define the secret name in this section and do not define peer.ca.secretName. |
| metadata    | Metadata of generated secret. |
| duration    | Expiration time of generated secret. |
| renewBefore | Time period before expiration time when certificate will be reissued. |
| privateKey  | Private key configuration: algorithm, key size and boolean parameter is it necessary to refresh private key when certificate is expired  |

### ClientServer section has the same fields as peer section.

### Rbac
| Option | Description |
| ------ | ----------- |
| enabled  | Enables role-based access control: creates root user in etcd, gives him root role and enables authentication in etcd. |

```
spec:
  security:
      peer:
        enabled: true # optional
        ca:
[...]

PR #125, opened 2024-04-06
Implement mounting user certificates and keys

Kirill-Garbar:
This PR allows to define user-managed peer, client-server certificates.
It implements security specification:

```yaml
spec:
...
  security:
    tls:
      peerTrustedCASecret: ca-peer-secret
      peerSecret: peer-secret
      serverSecret: server-secret
      clientTrustedCASecret: ca-client-secret
      clientSecret: client-secret
...
```

Enabled RBAC and root certificate watching **is not** part of this PR.


api/v1alpha1/etcdcluster_webhook.go
```diff
@@ -256,6 +266,47 @@ func (r *EtcdCluster) validatePdb() (admission.Warnings, field.ErrorList) {
...
+	}
+
+	security := r.Spec.Security
+
+	if security.Peer != nil {
+		if (security.Peer.Ca.SecretName != "" && security.Peer.Cert.SecretName == "") ||
+			(security.Peer.Ca.SecretName == "" && security.Peer.Cert.SecretName != "") {
+
+			allErrors = append(allErrors, field.Invalid(
+				field.NewPath("spec", "security", "peer.ca.secretName", "peer.cert.secretName"),
```
sircthulhu, line 284: Does it work correctly?

config/crd/bases/etcd.aenix.io_etcdclusters.yaml
sircthulhu, line 1: When updating `crd/bases` we need to update CRD in helm chart as well

api/v1alpha1/etcdcluster_webhook.go
Kirill-Garbar: What exactly works correctly and what means correctly? :) I checked and did not have any questions. However we can have a short call and I will show you.

config/crd/bases/etcd.aenix.io_etcdclusters.yaml
Kirill-Garbar: You are right. Will do. Probably we need to add a check to pipeline, because 100% somebody will forget. I will create a ticket.

sircthulhu: For now please use the command from #119 and update CRD in the helm chart. The file will be without extension in `tmp` folder

Kirill-Garbar: Good, ticket already exists. As I told in the previous comment - will do.

api/v1alpha1/etcdcluster_webhook.go
Kirill-Garbar: I rechecked and found that message was incorrect. See screenshot.
photo_2024-04-08 13 36 45

Fixed it.

examples/certificates/ca-certificate-peer.yaml
kvaps, line 1: Let's join them into the single file, containing all the certificates and EtcdCluster configured to use them and place it into:

```
examples/manifests/etcdcluster-with-external-certificates.yaml
```

Kirill-Garbar: Will do

Kirill-Garbar: Done

sircthulhu:
This PR relates to https://github.com/aenix-io/etcd-operator/issues/76

config/crd/bases/etcd.aenix.io_etcdclusters.yaml
Kirill-Garbar: Fixed

api/v1alpha1/etcdcluster_types.go
```diff
@@ -41,6 +41,9 @@ type EtcdClusterSpec struct {
 	// +optional
 	PodDisruptionBudgetTemplate *EmbeddedPodDisruptionBudget `json:"podDisruptionBudgetTemplate,omitempty"`
 	Storage                     StorageSpec                  `json:"storage"`
+	// Security describes security settings of etcd (authentication, certificates, rbac)
+	// +optional
+	Security *SecuritySpec `json:"security,omitempty"`
```
lllamnyp, line 46: This seems incosistent. Why should `Security` be a pointer to `SecuritySpec`, but `Storage` a `StorageSpec` literal?

api/v1alpha1/etcdcluster_types.go
```diff
@@ -200,6 +203,41 @@ type StorageSpec struct {
...
+	Cert SecretSpec `json:"cert,omitempty"`
+}
+
+type ClientServerSpec struct {
+	// +optional
+	Ca SecretSpec `json:"ca,omitempty"`
+	// +optional
+	ServerCert SecretSpec `json:"serverCert,omitempty"`
+	// +optional
+	RootClientCert SecretSpec `json:"rootClientCert,omitempty"`
```
lllamnyp, line 228: I think it would be better to adhere to etcd semantics (see flag names in the security section of the [configuration guide](https://etcd.io/docs/v3.5/op-guide/configuration/)). This is covered by the flag `--trusted-ca-file`. I suggest using a field name like `TrustedCA`.

internal/controller/factory/statefulset.go
```diff
@@ -161,6 +274,44 @@ func generateEtcdArgs(cluster *etcdaenixiov1alpha1.EtcdCluster) []string {
...
+	serverProtocol := "http"
+
+	clientTlsSettings := []string{"--auto-tls"}
+
+	if cluster.Spec.Security != nil && cluster.Spec.Security.ClientServer != nil {
+		clientTlsSettings = []string{
+			"--trusted-ca-file=/etc/etcd/pki/server/ca/ca.crt",
+			"--cert-file=/etc/etcd/pki/server/cert/tls.crt",
+			"--key-file=/etc/etcd/pki/server/cert/tls.key",
+			"--client-cert-auth=false",
```
lllamnyp, line 297: If `--trusted-ca-file` is not empty, it makes no sense to set `--client-cert-auth` to false, as `--trusted-ca-file` is NOT the CA, that signed `--cert-file`, but the CA that should sign client certificates when mTLS is enabled.

api/v1alpha1/etcdcluster_types.go
sergeyshevch: Looks like storage also should be pointer if we allow to create resource without specified storage

Kirill-Garbar: Storage was implemented some time ago - the first implementation. Also we have a bug in storage spec. So storage is not an example

internal/controller/factory/statefulset.go
Kirill-Garbar: Wow. This is bug. I missed it. It should not have =false. Thank you.

sergeyshevch (approved):

api/v1alpha1/etcdcluster_types.go
```diff
@@ -200,6 +203,41 @@ type StorageSpec struct {
...
+type PeerSpec struct {
+	// +optional
+	Ca SecretSpec `json:"ca,omitempty"`
+	// +optional
+	Cert SecretSpec `json:"cert,omitempty"`
+}
+
+type ClientServerSpec struct {
+	// +optional
+	Ca SecretSpec `json:"ca,omitempty"`
```
lllamnyp, line 224: This field needs a comment, documenting its purpose, e.g.
```go
  // Ca is the trusted CA used to sign client certificates...
  // +optional
  Ca SecretSpec `json:"ca,omitempty"`
  ```
  and in line with etcd's terminology `TrustedCA` might be a better name for it.

api/v1alpha1/etcdcluster_types.go
lllamnyp: Resolved in favor of the comment for lines 223-224.

api/v1alpha1/etcdcluster_types.go
Kirill-Garbar: We look here not only into etcd configuration, but also in the operator-managed certificates generation. That is why it is not only TRUSTED CA certificate, but also CA.KEY that signs client certificates.

For me TRUSTED means certificate without a key.

sircthulhu (approved):

kvaps (approved):

PR #165, opened 2024-04-12
Upd example with external certificates

kvaps:
Fix dnsNames and formating in etcdcluster-with-external-certificates.yaml


kvaps:
Closed due to misunderstood of useing -client service, see https://github.com/aenix-io/etcd-operator/issues/166

Issue #132, opened 2024-04-10
[API] not using core/v1 pod spec


Bo0km4n:
https://github.com/aenix-io/etcd-operator/blob/main/api/v1alpha1/etcdcluster_types.go#L140

etcd cluster pod template has  a member: "PodSpec". 
I seems like this field has same values as https://pkg.go.dev/k8s.io/api@v0.29.3/core/v1#PodSpec

Why doesn't use official package's struct, you defined by original?
If there is no specific reason, I think should import and use official package's struct

sergeyshevch:
Initially, the spec was inspired by the Victoriametrics operator and highly changed by the maintainers team. 

Our type gives us the ability to disable the propagation of some fields. Also, we can make sure that podSpec will work correctly in all K8s versions that the operator should support.

If you need some fields that are now not presented in our podSpec, please rise an issue and we will add it.

Bo0km4n:
I figured out your background, thank you.

The cluster controller create or update etcd pods from above PodSpec, so it should replace spec type to k8s core v1.
By doing so, you can notice new pod spec feature and deprecated it when update k8s dependencies.
I will contribute about this!

lllamnyp:
Thanks for your suggestion. The maintainers are divided in opinions on this matter. I would encourage you to offer your viewpoint under #172.

lllamnyp:
This is completed by #175. Let us know what you think and feel free to comment or open a new issue if you feel that there's more work to be done in this direction.

PR #167, opened 2024-04-12
Replace original PodSpec to corev1 PodSpec

Bo0km4n:
## Why
ref: #132 
These controllers create and update pod from template in crd. 
I think there is no reason wrap pod spec by original.

## What
- Use corev1 PodSpec
- Remove old PodSpec
- Fix tests


gecube:
hello @Bo0km4n 

I think that we intentionally used our version of PodSpec. The ratio is that we don't want to allow user to change ANY field in PodSpec. Let's imagine very simple example. We have image field on EtcdCluster spec level. And we have image on PodSpec level. What if both will be set?

Bo0km4n:
@gecube 
>I think that we intentionally used our version of PodSpec. 

I think there is no specific reason. please look #132 . 

> The ratio is that we don't want to allow user to change ANY field in PodSpec. 

What do you mean? 
It seems like implementation of etcdcluster webhook and controller don't restrict update pod spec of etcdcluster

>And we have image on PodSpec level. What if both will be set?

This commit just change PodSpec reference. etcdcluster have only one `image` filed, it is not duplicated.
I think this commit will not make breaking change for current user and resources.

sergeyshevch:
@Bo0km4n Let's wait for #172. For now, opinions in the maintainers team are different and we created a proposal to discuss it in more detail

lllamnyp:
This is done in #175.

PR #169, opened 2024-04-12
WIP: Refactoring certs and volumes for sts

sergeimonakhov:
This pull request introduces code changes related to managing volumes for the STS object. Key modifications include the addition of structures to store certificate information and the definition of paths to certificate files. Additionally, changes have been made to the creation and mounting of the data volume for the ETCD container. These alterations aim to enhance certificate handling and data management within the controller's context.

### DIFF
--- a/internal/controller/factory/statefulset.go
+++ b/internal/controller/factory/statefulset.go
@@ -109,6 +109,7 @@ func CreateOrUpdateStatefulSet(
 	}
 
 	volumes := generateVolumes(cluster)
+	containers := generateContainers(cluster)
 
 	statefulSet := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
@@ -126,7 +127,7 @@ func CreateOrUpdateStatefulSet(
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: podMetadata,
 				Spec: corev1.PodSpec{
-					Containers:                    generateContainers(cluster),
+					Containers:                    containers,
 					ImagePullSecrets:              cluster.Spec.PodTemplate.Spec.ImagePullSecrets,
 					Affinity:                      cluster.Spec.PodTemplate.Spec.Affinity,
 					NodeSelector:                  cluster.Spec.PodTemplate.Spec.NodeSelector,
@@ -352,18 +353,22 @@ func hasProbeHandlerAction(probe corev1.Probe) bool {
 	return probe.HTTPGet != nil || probe.TCPSocket != nil || probe.Exec != nil || probe.GRPC != nil
 }
 
+// Name returns the name of the certificate.
 func (c certificate) Name() string {
 	return c.certName
 }
 
+// MountPath returns the mount path associated with the certificate.
 func (c certificate) MountPath() string {
 	return c.mountPath
 }
 
+// CrtFilePath returns the file path of the certificate's CRT file.
 func (c certificate) CrtFilePath() string {
 	return c.buildFilePath(c.crtFile)
 }
 
+// KeyFilePath returns the file path of the certificate's key file, if available.
 func (c certificate) KeyFilePath() string {
 	if c.keyFile == "" {
 		return ""
@@ -371,10 +376,12 @@ func (c certificate) KeyFilePath() string {
 	return c.buildFilePath(c.keyFile)
 }
 
+// buildFilePath constructs the full file path using the certificate's mount path.
 func (c certificate) buildFilePath(fileName string) string {
 	return filepath.Join(c.mountPath, fileName)
 }
 
+// getCertificateConfig returns a certificate configuration object with the provided details.
 func getCertificateConfig(name, mountPath, crtFile, keyFile string) certificate {
 	return certificate{
 		certName:  name,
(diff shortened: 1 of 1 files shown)


sircthulhu:
@sergeimonakhov 
Please run `make lint-fix` to resolve `pre-commit` errors and fix tests

internal/controller/factory/statefulset.go
```diff
@@ -482,3 +374,269 @@ func mergeWithDefaultProbe(probe *corev1.Probe, defaultProbe corev1.Probe) *core
...
+		}
+	}
+
+	if !found {
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      volumeName,
+			ReadOnly:  readOnly,
+			MountPath: mountPath,
+		})
+	}
```
aobort, line 588: I'd suggest to use `slices`:

```go
if idx := slices.IndexFunc(volumeMounts, func(...) bool {...}; idx < 0 {
    volumeMounts = append(...)
}
```

internal/controller/factory/statefulset.go
```diff
@@ -482,3 +374,269 @@ func mergeWithDefaultProbe(probe *corev1.Probe, defaultProbe corev1.Probe) *core
...
+
+	return volumeMounts
+}
+
+func mergeVolumeMounts(lists ...[]corev1.VolumeMount) []corev1.VolumeMount {
+	var volumeMounts []corev1.VolumeMount
+
+	for _, list := range lists {
+		volumeMounts = append(volumeMounts, list...)
+	}
```
aobort, line 639: I'd suggest to use `copy`:

```go
volumeMounts := make([]corev1.VolumeMount, len(list))
copy(volumeMounts, list)
```

internal/controller/factory/statefulset.go
```diff
@@ -32,8 +33,62 @@ import (
...
+	DirName() string
+	DataPath() string
+}
+
+type etcd struct {
+	containerName string
+	volumeName    string
+	mountPath     string
+	dirName       string
+}
```
lllamnyp, line 65: IMO, it is premature to create a completely private struct and an interface wrapper around it. At the moment I don't understand the need for an abstraction like this, but an extra layer of indirection will make the code less readable.

Also I'm having trouble understanding the purpose of these structs. In our context `etcd` is quite a generic name. The fields/methods of these data structures hint at volumes, rather than an etcd cluster in general. A documentation comment would be quite helpful, as would a more specific name.

internal/controller/factory/statefulset.go
```diff
@@ -482,3 +374,269 @@ func mergeWithDefaultProbe(probe *corev1.Probe, defaultProbe corev1.Probe) *core
...
+	return e.mountPath
+}
+
+func (e etcd) DirName() string {
+	return e.dirName
+}
+
+func (e etcd) DataPath() string {
+	return filepath.Join(e.mountPath, e.dirName)
+}
```
lllamnyp, line 428: As a follow up to my previous comment, I would drop this for now and, rather than having functions, that accept an `etcdConfigurator` interface, I'd write functions that accept an `etcd` struct.

Compare

```go
func someFunc(ec etcdConfigurator) {
        doSomethingWith(ec.VolumeName())
}
```

and

```go
func someFunc(e etcd) {
        doSomethingWith(e.volumeName)
}
```

internal/controller/factory/statefulset.go
lllamnyp: A lot of this custom merging logic could be handled nicely with `StrategicMergePatch`.

internal/controller/factory/statefulset.go
lllamnyp: Or simply back on line 196
```go
volumeMounts := append(tlsVolumeMounts, dataVolumeMount...)
```

aobort: This function is used in the only place, function which accepts slice as an argument and merges this slice with computed slice in its body.
IMO, it makes no sense to write dedicated function for merging slices of concrete type which accepts variadic number of arguments if it is used in one place where the number of arguments which will be passed to that function is fixed.

aobort: > It won't work that way because the function mergeVolumeMounts accepts `...[]corev1.VolumeMount` as an argument, which allows passing N slices. While it's possible to calculate the total length of the slices, that would require a loop. In this case, using copy versus append would make no difference.

Yes, my bad. Missed that the argument is not a slice, but the slice of slices

aobort: > Initially, there were no certificates, but then they appeared. Now, let's say some config maps and who knows what else will appear later. How would the code look then? Right now, it's very simple: we just add slices to `volumeMounts := mergeVolumeMounts(tlsVolumeMounts, dataVolumeMount, configMapVolumeMount, etcVolumeMount)`.

In case of any volume mounts defined by the user, I would expect is that function `generateVolumeMounts` will accept as arguments:
1. predefined mounts - data, tls certs, whatnot else
1. container volumeMounts defined in cluster spec by the user

and merge these two lists. Computation of predefined mounts is the separate step where we need to take into account provided cluster configuration, like security config.
Bam, profit.

PR #175, opened 2024-04-15
Use StrategicMergePatch to apply overrides

lllamnyp:
This commit removes the custom etcdaenixiov1alpha1.PodSpec type in favor of corev1.PodSpec. As a result, StrategicMergePatch may be used to correctly applythe overrides from EtcdCluster.Spec.PodTemplate.Spec on top of the default PodSpec, generated by the operator. In turn, this allows removal of dozens of lines of code responsible for custom merging logic.


internal/controller/factory/statefulset.go
```diff
@@ -70,6 +71,28 @@ func CreateOrUpdateStatefulSet(
 
 	volumes := generateVolumes(cluster)
 
+	basePodSpec := corev1.PodSpec{
+		Containers: []corev1.Container{generateContainer(cluster)},
+		Volumes:    volumes,
+	}
+	basePodSpecBytes, err := json.Marshal(basePodSpec)
```
sergeyshevch, line 78: Looks like merge logic is pretty common and can be reused in multiple places. Maybe better to move to some helper

sergeyshevch (changes requested):

Makefile
```diff
@@ -49,6 +49,7 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	yq -i '.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.podTemplate.properties.spec.properties |= {}' config/crd/bases/etcd.aenix.io_etcdclusters.yaml
```
lllamnyp, line 52: I ran into the CRD being way too large as generated, so I borrowed the [approach from VictoriaMetrics](https://github.com/VictoriaMetrics/operator/blob/21ab04581f22d305d1e8265404fbfc5e4817ee25/Makefile#L71). Didn't manage to find a native way for kubebuilder to skip generation of a schema for nested fields of an embedded resource.

hiddenmarten: Please, use `yq` as `$(YQ)` and add `yq` as a dependency to the makefile target.

hiddenmarten: As an example:
https://github.com/aenix-io/etcd-operator/blob/4e3ddbe49ca00ae68bcab711627be60723019a0f/Makefile#L135-L141

hiddenmarten: Also, could you please provide a link or smth. about the issue itself. It's hard to understand what is the real problem...

lllamnyp: > Also, could you please provide a link or smth. about the issue itself. It's hard to understand what is the real problem...

Here's the [problem](https://book.kubebuilder.io/faq.html?highlight=maxdesclen#the-error-too-long-must-have-at-most-262144-bytes-is-faced-when-i-run-make-install-to-apply-the-crd-manifests-how-to-solve-it-why-this-error-is-faced). It's caused by a very large CRD spec being written into the `kubernetes.io/last-applied-configuration` annotation. The reason why the CRD is so large is because the entire `corev1.PodSpec` resource is being embedded into it.

kvaps: Could we use server-side-apply to avoid this situation?

lllamnyp: > Could we use server-side-apply to avoid this situation?

There's a bunch more problems rising from this. For one, there's artifacthub's [size limits for helm charts (see linked discussion in Russian)](https://t.me/VictoriaMetrics_ru1/52767). What's worse, we don't have much control over how users install our operator and the CRDs. **We** can do SSA, sure, but there's no way to make sure that our users will do SSA.

Kirill-Garbar: I am afraid to say it out loud. Should we redesign the spec then to avoid using podTemplate with k8s containers because it is too big?

kvaps: It sounds like a reasonable reason to abandon this idea.
I'm generally not opposed to adding fields to the spec when it is really necessary.

However, I would like to have a common logic for merging.

hiddenmarten (changes requested):
Update Makefile as an example from the thread.

internal/controller/factory/statefulset.go
lllamnyp: I would like to break this out into a separate function, but once I do, it immediately needs coverage with unit tests. How about we leave it as is and, once (if) we merge, I'll make an issue to factor it out and write some tests?

internal/controller/factory/statefulset.go
```diff
@@ -70,6 +71,31 @@ func CreateOrUpdateStatefulSet(
...
+	if err != nil {
+		return fmt.Errorf("cannot marshal generated base pod spec to JSON: %w", err)
+	}
+	if cluster.Spec.PodTemplate.Spec.Containers == nil {
+		cluster.Spec.PodTemplate.Spec.Containers = make([]corev1.Container, 0)
+	}
```
lllamnyp, line 84: It so happens that the JSON marshal of an empty PodSpec is not `{}`. but `{"containers":null}` (the only field without omitEmpty, I guess). So I have to explicitly set it to an empty list, or else `StrategicMergePatch` deletes the `containers` field in the final statefulset.

Makefile
lllamnyp: With the post-processing, as presented in the diff to the Makefile, this becomes a non-issue. There's also no problem in taking the contents of the `PodTemplate` field and validating them according to the correct schema in a validating webhook.

hiddenmarten (approved):

gecube (approved):

sergeyshevch (approved):

Natthachai0434:
eaca8d6006a18a05ffe8dfe7009739261d21b0bd

PR #184, opened 2024-04-20
Add nilaway linter.

### LINKED ISSUE #174: Nilaway, check in pre-commit
Add https://github.com/uber-go/nilaway as a plugin for golanci-lint.

https://github.com/uber-go/nilaway?tab=readme-ov-file#golangci-lint

mneverov:
Closes #174.

Without specifying packages the linter consume all memory when trying to parse `site` and other directories.
I excluded `cmd` package because the linter complained on using [mgr](https://github.com/aenix-io/etcd-operator/blob/main/cmd/main.go#L126) even if it cannot be nil since the error that is returned from the manager constructer is checked.


hiddenmarten:
Hi!

Thank you for your contribution. I'm gonna review this PR.

Makefile
```diff
@@ -270,6 +275,10 @@ golangci-lint: $(LOCALBIN)
 	@test -x $(GOLANGCI_LINT) && $(GOLANGCI_LINT) version | grep -q $(GOLANGCI_LINT_VERSION) || \
 	GOBIN=$(LOCALBIN) go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
 
+.PHONY: nilaway
+nilaway: $(LOCALBIN)
+	@test -x $(NILAWAY_LINT) || GOBIN=$(LOCALBIN) go install go.uber.org/nilaway/cmd/nilaway@latest
```
hiddenmarten, line 280: I've checked it.
It's so sad that we can't use particular version ;(

hiddenmarten (approved):

Makefile
mneverov: yes, they don't have releases

sircthulhu (approved):

PR #217, opened 2024-05-19
feat: set quota-backend-bytes to 95% of requested size

### LINKED ISSUE #211: Automatically set `quota-backend-bytes` option to 95% from requested size:
eg.:
```
spec:
  storage:
    volumeClaimTemplate:
      metadata: {}
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 20Gi
```

should also implicitly set:

```
spec:
  options:
    quota-backend-bytes: "20401094656"
```

also we have to check if this option not applied by user explicitly

sircthulhu:
- Changed `ENVTEST_K8S_VERSION` to `1.30.0` as there's no tool chains for version `1.30.1`
- Implemented auto setting `quota-backend-bytes` for requested PVC / emptyDir size

Possible problems:
- After changing requested size, option `quota-backend-bytes` will be left unchanged

fixes #211

### DIFF
--- a/internal/controller/factory/statefulset.go
+++ b/internal/controller/factory/statefulset.go
@@ -36,7 +36,8 @@ import (
 )
 
 const (
-	etcdContainerName = "etcd"
+	etcdContainerName                = "etcd"
+	defaultBackendQuotaBytesFraction = 0.95
 )
 
 func CreateOrUpdateStatefulSet(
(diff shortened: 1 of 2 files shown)


internal/controller/factory/statefulset.go
```diff
@@ -247,6 +250,25 @@ func generateEtcdCommand() []string {
...
+	if value, ok := cluster.Spec.Options["quota-backend-bytes"]; !ok || value == "" {
+		var size resource.Quantity
+		if cluster.Spec.Storage.EmptyDir != nil {
+			if cluster.Spec.Storage.EmptyDir.SizeLimit != nil {
+				size = *cluster.Spec.Storage.EmptyDir.SizeLimit
+			}
+		} else {
+			size = *cluster.Spec.Storage.VolumeClaimTemplate.Spec.Resources.Requests.Storage()
+		}
+		quota := float64(size.Value()) * 0.95
```
sergeyshevch, line 262: Maybe we need to make this constant configurable somewhere

sergeyshevch (approved):

Issue #208, opened 2024-05-14
Disable development logger


kvaps:
https://github.com/aenix-io/etcd-operator/blob/8157aa3601021bfc748c0f10bef4606494a4aa29/cmd/main.go#L69-L71

PR #227, opened 2024-05-28
update logging

aobort:
Changes:
- explicit flags to set logging level, logging level to attach stacktrace, development mode
- explicit flag to disable webhooks
- logging wrapper with zap & slog (might be reasonable to replace it with custom slog handler and remove zap in the future)
- replace `logr.V(N).*` calls with more readable and clear wrapped `log.Info(...)`, `log.Debug(...)` and `log.Error(...)`

Fixes: #208


kvaps (approved):
LGTM

internal/controller/factory/pdb.go
```diff
@@ -21,49 +21,49 @@ import (
...
+	ctx, err := contextWithGVK(ctx, pdb, rclient.Scheme())
+	if err != nil {
+		return err
+	}
+
+	if cluster.Spec.PodDisruptionBudgetTemplate == nil {
+		return deleteOwnedResource(ctx, rclient, pdb)
+	}
+
+	pdb.Spec = v1.PodDisruptionBudgetSpec{
```
Kirill-Garbar, line 53: As I understood you restructured this piece of code a bit to avoid duplication of 
```
ObjectMeta: metav1.ObjectMeta{
				Namespace: cluster.Namespace,
				Name:      cluster.Name,
			}
```

Is it possible to achieve the same without separation of pdb object to ObjectMeta and pdb.Spec?

The problem is that when such object structures are separated in the code, in the end it is sometimes more complicated to search them in different pieces of code. In this case they are close to each other, but in the future they can go away from each other and be placed in different pieces of the go file. Readability suffers in my opinion.

From my perspective always easier to have resulted object describe in one place, than combine objects like here: https://github.com/aenix-io/etcd-operator/blob/010228b82e26a605f78449a5fbfd7350f0fd6475/internal/controller/factory/statefulset.go#L358

Kirill-Garbar (approved):

Kirill-Garbar:
Also could you please add comments to functions and methods? For now it is not clear from the first glance what are the functions for.

internal/controller/factory/pdb.go
aobort: @Kirill-Garbar totally agree with you. Unfortunately, I didn't find a way how to do it without making this piece of code even more ugly. The only option which might be suitable from my perspective is to manually set context values in case of pdb deletion:
```go
	if cluster.Spec.PodDisruptionBudgetTemplate == nil {
		ctx = log.WithValues(ctx, "group", "v1", "kind", "PodDisruptionBudget", "name", cluster.Name)
		return deleteOwnedResource(ctx, rclient, &v1.PodDisruptionBudget{
			ObjectMeta: metav1.ObjectMeta{
				Namespace: cluster.Namespace,
				Name:      cluster.Name,
			},
		})
	}
```
Not the best solution, but still. WDYT?

Kirill-Garbar: I think it's fine. I do not have better idea.

PR #229, opened 2024-06-01
[222] Add autocomaction to ConfigMap

hiddenmarten:


sircthulhu (approved):

internal/controller/factory/configMap.go
```diff
@@ -56,9 +56,11 @@ func CreateOrUpdateClusterStateConfigMap(
...
 		},
 		Data: map[string]string{
-			"ETCD_INITIAL_CLUSTER_STATE": "new",
-			"ETCD_INITIAL_CLUSTER":       initialCluster,
-			"ETCD_INITIAL_CLUSTER_TOKEN": cluster.Name + "-" + cluster.Namespace,
+			"ETCD_INITIAL_CLUSTER_STATE":     "new",
+			"ETCD_INITIAL_CLUSTER":           initialCluster,
+			"ETCD_INITIAL_CLUSTER_TOKEN":     cluster.Name + "-" + cluster.Namespace,
+			"ETCD_AUTO_COMPACTION_RETENTION": "5m",
+			"ETCD_SNAPSHOT_COUNT":            "10000",
```
Kirill-Garbar, line 63: I think clusterState config map isn't supposed to have different etcd settings.

hiddenmarten: It makes sense, I've updated the configmap name.

Kirill-Garbar: We discussed in telegram chat the way how to operate cluster state information. This configmap is supposed to be filled in by the operator after getting info from the etcd cluster. And supposed to stay cluster state configmap.

Let's leave it cluster state cm and not rename it and not fill in with anything else.

We have function where all options are handled (set by the user as additional ones and set by operator): https://github.com/aenix-io/etcd-operator/blob/main/internal/controller/factory/statefulset.go#L251

Kirill-Garbar (approved):

PR #231, opened 2024-06-04
add resizing operation

sevagol:
added a comparison of pvc size with the existing one, and deletion in case of discrepancy. Should a new statefulset be created automatically after this?


kvaps:
https://github.com/aenix-io/etcd-operator/blob/010228b82e26a605f78449a5fbfd7350f0fd6475/internal/controller/etcdcluster_controller.go#L136

internal/controller/etcdcluster_controller.go
```diff
@@ -67,22 +70,64 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			err = r.Get(ctx, client.ObjectKey{Name: pvcName, Namespace: instance.Namespace}, pvc)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					logger.V(2).Info("PVC not found, skipping storage check", "pvc_name", pvcName)
+				} else {
+					return reconcile.Result{}, err
+				}
+			} else {
+				desiredStorage := instance.Spec.Storage.VolumeClaimTemplate.Spec.Resources.Requests[corev1.ResourceStorage]
+				currentStorage := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
```
kvaps, line 108: you have to compare storage from `statefulSet.Spec.VolumeClaimTemplate`, not pvc

kvaps (commented):
Please add this functionaity before this function

https://github.com/aenix-io/etcd-operator/blob/010228b82e26a605f78449a5fbfd7350f0fd6475/internal/controller/etcdcluster_controller.go#L136

It should be separate function eg.:

```
if isStatefulSetDeletionRequired; {
    if err := factory.DeleteStatefulSet(ctx, cluster, r.Client); err != nil {
    	return err
    }
}
```

makhov:
Do I understand it properly: with this PR we just update the StatefulSet without actual resizing of the PVCs? 

Since the PVCs are already created they won't be changed even if the `volumeClaimTemplates` are changed. Here is the doc on how to resize the PVC:
https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

kvaps:
Cirrect, the resizing of PVCs will probably implemented in separate PR

kvaps:
closed if favor https://github.com/aenix-io/etcd-operator/pull/254

Issue #235, opened 2024-06-12
Fix logging system issues


Kirill-Garbar:
- Previously content of k8s secrets wasn't logged. After logging system change content is logged.
- Nil pointer exception panic if context without logger (context.Background()) is passed.

PR #246, opened 2024-06-30
logging enhancement

aobort:
Added:
- stacktrace to log entry
- cmd flag to enable/disable stacktrace (by default stacktrace enabled, stacktrace log level = error)

To disable stacktrace in logs, one should explicitly provide `--enable-stacktrace=false` flag or set corresponding env var (`ENABLE_STACKTRACE=false`).

Fixed:
- fallback to default logger, if context does not contain logger
- sensitive data in logged tls v1.Secret is now omitted

Fixes: #235


sircthulhu (approved):
lgtm

kvaps (approved):

PR #250, opened 2024-07-10
Set extraArgs in the end of args processing

makhov:
Set extraArgs in the end of args processing to let them overwrite default values

### DIFF
--- a/internal/controller/factory/statefulset_test.go
+++ b/internal/controller/factory/statefulset_test.go
@@ -17,8 +17,6 @@ limitations under the License.
 package factory
 
 import (
-	"slices"
-
 	"github.com/google/uuid"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


internal/controller/factory/statefulset_test.go
```diff
@@ -180,11 +178,6 @@ var _ = Describe("CreateOrUpdateStatefulSet handler", func() {
 
 			By("Checking the extraArgs", func() {
 				Expect(statefulSet.Spec.Template.Spec.Containers[0].Args).To(Equal(generateEtcdArgs(&etcdcluster)))
-				By("Checking args are sorted", func() {
```
makhov, line 183: They are not all sorted anymore, so remove the test

sircthulhu: I think tests should check that extraArgs in `statefulSet.Spec.Template.Spec.Containers[0].Args` are sorted. So that pods won't be recreated on the next reconciliation

makhov: Currently, it doesn't make much sense since `spec.Options` field is empty, so we will sort a slice with only one element.

makhov: Updated the test, but I don't really like the approach :)

sircthulhu (approved):
lgtm

Kirill-Garbar (approved):

PR #207, opened 2024-05-12
Extended status check for reconciliation

lllamnyp:
This draft PR is an implementation of #181

PR #181, opened 2024-04-18
Create a design-document for the controller

lllamnyp:
# Motivation

I started some "R'n'D" (scare quotes intended) for implementing scale up, scale down, self-healing and so on and quickly realized, that the coding of the member add/member remove and similar steps is the more trivial part of the undertaking. The difficult part is  coming up with a working algorithm that can correctly deduce the cluster's state and execute the necessary actions at the right time.

To better reason about the controller's algorithm now, and to better develop it going forward, I feel it is important to have good documentation of the current design and the intended next steps, so I started with trying to document the current state of the code.

# Results

This document contains a mermaid flowchart that outlines the reconciliation loop. It is better viewed in [rendered form](https://github.com/aenix-io/etcd-operator/blob/docs/design/docs/DESIGN.md).

Going forward, I envision this document to have at least three purposes:
* Let the developers spot flaws and prompt them to open issues.
* Act as a more detailed form of documentation for advanced users.
* Be a blueprint for implementing anything non-trivial.

## Summary by CodeRabbit

- **Documentation**
	- Updated the design document for the `EtcdCluster` custom resources with a detailed flowchart illustrating the reconciliation process and lifecycle management within a Kubernetes environment.

PR #252, opened 2024-07-11
Add a helper set library to facilitate implementation of the new health check procedure

lllamnyp:
This is the first PR in a series of [stacked PRs](https://www.stacking.dev/), aimed ultimately at implementing the features described in #181 and #207. The next PR in the stack can be found at #259.

### DIFF
--- a/pkg/set/set.go
+++ b/pkg/set/set.go
@@ -1,0 +1,29 @@
+package set
+
+type Set[E comparable] map[E]struct{}
+
+func New[E comparable](vals ...E) Set[E] {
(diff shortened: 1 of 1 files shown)


pkg/set/set.go
```diff
@@ -0,0 +1,29 @@
+package set
```
makhov, line 1: Just a note: there is a [map](https://pkg.go.dev/maps) package in the standard library, that may cover your needs

kvaps (approved):
LGTM

sircthulhu (approved):

Issue #210, opened 2024-05-15
Enable resize operation


kvaps:
**phase1:**

if .spec.storage.volumeClaimTemplate.spec.resources.requestes.storage updated:

```
  storage:
    volumeClaimTemplate:
      metadata: {}
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 20Gi
```

then delete statefulset with `--cascade=orphan` and create a new one

**phase2:**

Loop over all PVCs

if storageClass has `allowVolumeExpansion: true`

then patch PVC object to higher size
```
spec:
  resources:
    requests:
      storage: 20Gi
```

PR #254, opened 2024-07-18
Add resizing operation

kvaps:
fixes https://github.com/aenix-io/etcd-operator/issues/210

Signed-off-by: Andrei Kvapil <kvapss@gmail.com>

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -57,6 +57,21 @@ rules:
   - update
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - "storage.k8s.io"
+  resources:
+  - storageclasses
+  verbs:
+  - get
+- apiGroups:
   - etcd.aenix.io
   resources:
   - etcdclusters
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -27,6 +27,15 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - get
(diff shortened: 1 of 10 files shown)


config/rbac/role.yaml
```diff
@@ -56,6 +56,21 @@ rules:
...
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - "storage.k8s.io"
+  resources:
+  - storageclasses
+  verbs:
+  - get
```
aobort, line 73: Pls, add the following in above of the `func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error)` (https://github.com/aenix-io/etcd-operator/blob/main/internal/controller/etcdcluster_controller.go#L72) to generate `role.yaml`:

```go
// +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=get;list;patch;watch
// +kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get
```

Otherwise it will be overwritten by the next 'make manifests' run.

aobort (changes requested):

config/rbac/role.yaml
kvaps: sure, done

aobort (approved):
lgtm

internal/controller/etcdcluster_controller.go
```diff
@@ -199,6 +207,28 @@ func (r *EtcdClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
+// checkAndDeleteStatefulSetIfNecessary deletes the StatefulSet if the specified storage size has changed.
+func (r *EtcdClusterReconciler) checkAndDeleteStatefulSetIfNecessary(ctx context.Context, state *observables, instance *etcdaenixiov1alpha1.EtcdCluster) error {
+	for _, volumeClaimTemplate := range state.statefulSet.Spec.VolumeClaimTemplates {
+		if volumeClaimTemplate.Name != "data" {
+			continue
+		}
+		currentStorage := volumeClaimTemplate.Spec.Resources.Requests[corev1.ResourceStorage]
+		desiredStorage := instance.Spec.Storage.VolumeClaimTemplate.Spec.Resources.Requests[corev1.ResourceStorage]
+		if desiredStorage.Cmp(currentStorage) != 0 {
```
sircthulhu, line 218: I think we should check in validating webhook that new desired size is greater or equal to previous

kvaps: sure, done

sircthulhu (approved):

PR #259, opened 2024-08-12
Separate ConfigMap templating and creation

lllamnyp:
This PR is part of a stacked pull request on top of #252. The big picture is documented in #181 and #207.

This commit factors out the building of a corev1.ConfigMap struct into a separate function, while leaving the logic of determining key parameters, such as the initial cluster, state, etc, as well as sending the request to the k8s API server in the already existing CreateOrUpdate function. This will be useful in the future, when this logic will be moved out of the factory library and into the reconciliation loop.

## Summary by CodeRabbit

- **New Features**
	- Improved modularity and clarity in cluster state management with the introduction of a new helper function for creating configuration maps.
  
- **Refactor**
	- Simplified the `CreateOrUpdateClusterStateConfigMap` function for better readability and maintainability by delegating the configuration map creation.

### DIFF
--- a/internal/controller/factory/configmap.go
+++ b/internal/controller/factory/configmap.go
@@ -19,6 +19,7 @@ package factory
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	etcdaenixiov1alpha1 "github.com/aenix-io/etcd-operator/api/v1alpha1"
 	"github.com/aenix-io/etcd-operator/internal/log"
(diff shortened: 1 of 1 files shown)


internal/controller/factory/configmap.go
```diff
@@ -39,39 +40,16 @@ func CreateOrUpdateClusterStateConfigMap(
...
-			Name:      GetClusterStateConfigMapName(cluster),
-		},
-		Data: map[string]string{
-			"ETCD_INITIAL_CLUSTER_STATE": "new",
-			"ETCD_INITIAL_CLUSTER":       initialCluster,
-			"ETCD_INITIAL_CLUSTER_TOKEN": cluster.Name + "-" + cluster.Namespace,
-		},
+	state := "new"
+	if isEtcdClusterReady(cluster) {
+		state = "existing"
```
sircthulhu, line 45: Do we need deleted log line `log.Debug(ctx, "updating cluster state")` ?

sircthulhu (approved):

Kirill-Garbar (approved):

internal/controller/factory/configmap.go
lllamnyp: IMO, logging changes stored in an in-memory variable is excessive. This is essentially

```go
log.Debug("incrementing a by 1")
a++
```

The correct approach would be

```go
a++
log.Debug("updating a")
client.Update(&a)
```

PR #301, opened 2025-12-03
feat: remove empty properties from etcd-cluster spec to avoid ArgoCD drifting

apenen:
ArgoCD keeps detecting pending changes because Kubernetes omits empty fields on apply. Removing these fields from the spec prevents unnecessary diffs.

## Summary by CodeRabbit

* **Schema Updates**
  * Updated the etcd-operator CRD: validation for pod template specification has been relaxed to accept a broader set of fields while retaining the requirement for container definitions and preserving unknown fields. This enables more flexible pod template configurations without changing required structural elements.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


sircthulhu:
Hi, @apenen! This change will be overriden by commands
```bash
make manifests
make helm-crd-copy
```
So changes need to be made to `api/v1alpha1/etcdcluster_types.go`

PR #307, opened 2026-03-07
feat(backup): add EtcdBackup CRD for declarative etcd snapshots

IvanHunters:
## Summary

- Add `EtcdBackup` CRD — a one-shot resource that triggers a backup Job to take an etcd snapshot and store it to S3-compatible storage or a PVC
- Add `backup-agent` binary (built into the same image) that connects to etcd, takes a snapshot, and uploads it to the configured destination
- Add validating webhook enforcing immutable spec, exactly-one-destination constraint, and subPath path traversal protection
- Add Helm chart support: RBAC rules, webhook config, `OPERATOR_IMAGE` env injection, CRD

## Design

```text
User creates EtcdBackup CR
        |
EtcdBackupReconciler reads EtcdBackup + referenced EtcdCluster
        |
Controller creates a Job (backup-agent container)
        |
backup-agent connects to etcd, takes snapshot, uploads to S3/PVC
        |
Controller watches Job status -> updates EtcdBackup conditions
```

## New files

| File | Purpose |
| --- | --- |
| `api/v1alpha1/etcdbackup_types.go` | CRD types (spec, status, conditions) |
| `api/v1alpha1/etcdbackup_webhook.go` | Validating webhook |
| `cmd/backup-agent/main.go` | Backup agent binary |
| `internal/controller/etcdbackup_controller.go` | Reconciler |
| `internal/controller/factory/backup_job.go` | Job factory |

## Test plan

- [ ] Webhook tests: valid S3/PVC, both set, neither set, empty fields, subPath traversal, spec immutability, status-only updates allowed
- [ ] Controller tests: Job creation (PVC/S3), cluster not found, Job success/failure conditions, already complete skip, TLS volume mounts
- [ ] Factory tests: PVC/S3 job specs, TLS volumes, SubPath handling, labels, owner references
- [ ] `make generate && make manifests` produce clean output
- [ ] `make test` passes all suites
- [ ] `make docker-build` builds both binaries

## Summary by CodeRabbit

* **New Features**
  * EtcdBackup and EtcdBackupSchedule CRDs with S3/PVC destinations; backup and scheduled backup orchestration; restore-at-bootstrap support; backup-agent and restore-agent binaries; operator image propagated to workloads.
* **Chores**
  * Builds now produce manager, backup-agent, and restore-agent; charts, CRD manifests, samples, and packaging updated.
* **Security & Ops**
  * Validating webhooks added, cert-manager CA injection, and expanded RBAC for backups, schedules, jobs, and cronjobs.
* **Tests**
  * Extensive unit and integration tests covering webhooks, controllers, factory helpers, agents, jobs, cronjobs, and restore flows.


charts/etcd-operator/templates/workload/deployment.yml
```diff
@@ -66,6 +66,8 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.namespace
+          - name: OPERATOR_IMAGE
+            value: {{ .Values.etcdOperator.image.repository }}:{{ .Values.etcdOperator.image.tag | default .Chart.AppVersion }}
```
sircthulhu, line 70: This may be done with fieldPath

cmd/backup-agent/main.go
```diff
@@ -0,0 +1,205 @@
...
+	cfg := aws.Config{
+		Region:      region,
+		Credentials: credentials.NewStaticCredentialsProvider(accessKey, secretKey, ""),
+	}
+
+	var s3Client *s3.Client
+	if endpoint != "" {
+		s3Client = s3.NewFromConfig(cfg, func(o *s3.Options) {
+			o.BaseEndpoint = aws.String(endpoint)
+			o.UsePathStyle = true
```
sircthulhu, line 225: This won't work if the specified bucket uses vhost style

internal/controller/factory/backup_job.go
```diff
@@ -0,0 +1,203 @@
...
+						LocalObjectReference: corev1.LocalObjectReference{Name: s3.CredentialsSecretRef.Name},
+						Key:                  "AWS_SECRET_ACCESS_KEY",
+					},
+				},
+			},
+		)
+	}
+
+	if pvc := backup.Spec.Destination.PVC; pvc != nil {
+		backupPath := "/backup/data/snapshot.db"
```
sircthulhu: Hardcoding such path might lead to errors. Moreover, it does not include `EtcdBackup` resource name, so every next backup will overwrite previous

sircthulhu (changes requested):

internal/controller/factory/backup_job.go
```diff
@@ -0,0 +1,236 @@
...
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: corev1.PodSpec{
+					RestartPolicy: corev1.RestartPolicyNever,
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: ptr.To(true),
+						RunAsUser:    ptr.To(int64(65532)),
+						RunAsGroup:   ptr.To(int64(65532)),
```
sircthulhu, line 102: Missing `FSGroup` — the pod runs as UID 65532, but PVC mounts default to root ownership on many storage classes. Without `FSGroup: ptr.To(int64(65532))`, `writeToPVC()` will get `permission denied`.

Same applies to `backup_cronjob.go`.

For reference, Percona XtraDB operator always sets `FSGroup` in backup Job specs.

internal/controller/etcdbackup_controller.go
```diff
@@ -0,0 +1,196 @@
...
+			return r.updateStatus(ctx, backup)
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get EtcdCluster: %w", err)
+	}
+
+	jobName := factory.GetBackupJobName(backup)
+	existingJob := &batchv1.Job{}
+	err := r.Get(ctx, types.NamespacedName{Name: jobName, Namespace: backup.Namespace}, existingJob)
+	if err == nil {
+		return r.reconcileJobStatus(ctx, backup, existingJob)
```
sircthulhu: No ownership verification. A leftover Job from a previously deleted EtcdBackup with the same name, or a manually created Job, will be trusted blindly.

```go
if err == nil {
    if !metav1.IsControlledBy(existingJob, backup) {
        meta.SetStatusCondition(&backup.Status.Conditions, metav1.Condition{
            Type:               etcdaenixiov1alpha1.EtcdBackupConditionFailed,
            Status:             metav1.ConditionTrue,
            Reason:             "JobNameConflict",
            Message:            fmt.Sprintf("Job %q already exists and is not controlled by this EtcdBackup", jobName),
            ObservedGeneration: backup.Generation,
        })
        return r.updateStatus(ctx, backup)
    }
    return r.reconcileJobStatus(ctx, backup, existingJob)
}
```

internal/controller/etcdbackupschedule_controller.go
```diff
@@ -0,0 +1,260 @@
...
+// avoiding false positives from API Server-defaulted fields like
+// TerminationMessagePath, ImagePullPolicy, etc.
+func cronJobNeedsUpdate(existing, desired *batchv1.CronJob) bool {
+	if existing.Spec.Schedule != desired.Spec.Schedule {
+		return true
+	}
+	if !reflect.DeepEqual(existing.Spec.SuccessfulJobsHistoryLimit, desired.Spec.SuccessfulJobsHistoryLimit) {
+		return true
+	}
+	if !reflect.DeepEqual(existing.Spec.FailedJobsHistoryLimit, desired.Spec.FailedJobsHistoryLimit) {
```
sircthulhu: `ConcurrencyPolicy` is not compared here. The factory always sets `ForbidConcurrent`, but if someone changes it via kubectl, the drift will not be corrected.

```go
if existing.Spec.ConcurrencyPolicy != desired.Spec.ConcurrencyPolicy {
    return true
}
```

api/v1alpha1/etcdbackup_webhook.go
```diff
@@ -0,0 +1,178 @@
...
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *EtcdBackup) ValidateCreate() (admission.Warnings, error) {
+	etcdbackuplog.Info("validate create", "name", r.Name)
+
+	var allErrors field.ErrorList
+
+	// Job name = "{name}-backup" (7 char suffix).
+	// Job names must be <= 52 chars (63 max Pod name - 11 for Job/Pod suffixes).
+	const jobSuffix = "-backup"
+	const maxJobNameLen = 52
```
sircthulhu: This applies the CronJob name limit (52) to a plain Job. Job names can be up to 63 characters. The 52-char limit exists for CronJobs because the CronJob controller appends an 11-char suffix when creating child Jobs.

Since `GetBackupJobName()` creates a Job (not a CronJob), use:
```go
const maxJobNameLen = 63
```

config/crd/patches/webhook_in_etcdbackups.yaml
```diff
@@ -0,0 +1,16 @@
+# The following patch enables a conversion webhook for the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: etcdbackups.etcd.aenix.io
+spec:
+  conversion:
+    strategy: Webhook
```
sircthulhu: This enables `conversion.strategy: Webhook` but there are no conversion handlers (`ConvertTo`/`ConvertFrom`) and only one API version (`v1alpha1`) exists. The API server will fail when attempting to call the non-existent `/convert` endpoint.

Remove the conversion webhook configuration or set strategy to `None`.

config/samples/etcd.aenix.io_v1alpha1_etcdcluster_restore.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: etcd.aenix.io/v1alpha1
+kind: EtcdCluster
+metadata:
+  name: restored-cluster
+spec:
+  replicas: 3
```
sircthulhu: Sample uses `replicas: 3` with PVC restore. All StatefulSet pods will try to mount the same PVC. With `ReadWriteOnce` (the most common access mode), scheduling will stall for pods on different nodes.

Either change to `replicas: 1` with a comment about scaling up after restore, or use an S3 example instead.

For reference:
- **CloudNativePG** restores to a single primary first, then uses streaming replication
- **Percona** uses lease-based locking with `AllowParallel=false`

internal/controller/factory/statefulset.go
```diff
@@ -439,6 +451,167 @@ func getLivenessProbe() *corev1.Probe {
...
+
+	var extraVolumes []corev1.Volume
+
+	if s3 := restore.Source.S3; s3 != nil {
+		forcePathStyle := "false"
+		if s3.ForcePathStyle {
+			forcePathStyle = "true"
+		}
+		restoreAgentEnv = append(restoreAgentEnv,
+			corev1.EnvVar{Name: "RESTORE_SOURCE", Value: "s3"},
```
sircthulhu, line 493: S3/PVC env var construction is duplicated here and in `buildBackupContainer()` (`backup_job.go`). Consider extracting a shared helper like `buildDestinationEnvVars(dest BackupDestination) ([]corev1.EnvVar, []corev1.Volume, []corev1.VolumeMount)` to avoid divergence when adding new fields to `BackupDestination`.

internal/controller/factory/backup_job.go
```diff
@@ -0,0 +1,236 @@
...
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      GetBackupJobName(backup),
+			Namespace: backup.Namespace,
+			Labels:    labels,
+		},
+		Spec: batchv1.JobSpec{
+			BackoffLimit:            &backoffLimit,
+			TTLSecondsAfterFinished: &ttl,
```
sircthulhu, line 91: Two suggestions:

1. **Add `ActiveDeadlineSeconds`** to the Job spec. `BACKUP_TIMEOUT_MINUTES` only works inside the process — if backup-agent hangs on a syscall (NFS, unresponsive S3), the internal timeout will not fire. A Kubernetes-level deadline is the safety net.

2. **Add `ephemeral-storage`** to resource requests for S3 backups — the snapshot is written to a temp file before upload, and without resource accounting the pod can be evicted on nodes with ephemeral storage limits.

internal/controller/factory/backup_job.go
```diff
@@ -0,0 +1,236 @@
...
+	if s3 := destination.S3; s3 != nil {
+		forcePathStyle := "false"
+		if s3.ForcePathStyle {
+			forcePathStyle = "true"
+		}
+		envVars = append(envVars,
+			corev1.EnvVar{Name: "BACKUP_DESTINATION", Value: "s3"},
+			corev1.EnvVar{Name: "S3_ENDPOINT", Value: s3.Endpoint},
+			corev1.EnvVar{Name: "S3_BUCKET", Value: s3.Bucket},
+			corev1.EnvVar{Name: "S3_KEY", Value: s3.Key},
```
sircthulhu: **Backup file naming should be the operator's responsibility, not the user's.**

Currently the naming logic is inconsistent:
- **PVC one-shot**: `/backup/data/{backupName}.db` (good, but `SubPath` overrides this completely)
- **PVC scheduled**: `/backup/data/{scheduleName}.db` + timestamp injected inside backup-agent (naming logic split between factory and agent)
- **S3 one-shot**: user specifies full key — two EtcdBackup resources with the same `key` silently overwrite each other
- **S3 scheduled**: same key + timestamp injected by backup-agent

Proposed approach: the user should specify a **directory/prefix**, and the operator should construct the full path deterministically:
- One-shot EtcdBackup: `{prefix}/{etcdbackup-name}.db`
- EtcdBackupSchedule: `{prefix}/{etcdbackupschedule-name}-{YYYYMMDD-HHmmSS}.db`

This keeps naming logic in one place (the factory), makes backup artifacts traceable to their source resources, and prevents silent overwrites.

For reference:
- **coreos/etcd-operator** appends `_v{revision}_{timestamp}` to the base path
- **Percona** uses the backup resource name as the S3 key prefix: `s3://{bucket}/{backupName}`
- **CloudNativePG** generates deterministic names: `{scheduledBackupName}-{compactISO8601}`

sircthulhu (dismissed):
## Review Summary

Good architecture overall — two CRDs, Job-based execution, CronJob for scheduling. This aligns well with patterns from CloudNativePG and Percona operators.

I've compared the implementation with coreos/etcd-operator, CloudNativePG, and Percona XtraDB Cluster Operator. Key findings are in the inline comments below.

### Critical
1. Missing `FSGroup` in PodSecurityContext — PVC writes will fail on most storage classes
2. Conversion webhook patch is invalid — no conversion handlers exist
3. No ownership check when adopting existing Job
4. Backup file naming — operator should own naming, user should only specify directory/prefix

### Important
5. `cronJobNeedsUpdate` doesn't compare `ConcurrencyPolicy`
6. No `ephemeral-storage` resource requests for S3 temp file
7. PVC restore sample uses `replicas: 3` with potentially `ReadWriteOnce` PVC
8. Job name length limit uses 52 (CronJob limit) instead of 63 (Job limit) for EtcdBackup

### Suggestions
9. S3/PVC env var generation is duplicated between `backup_job.go` and `statefulset.go`
10. No `activeDeadlineSeconds` on backup Job
11. TTL of 600s is aggressive for debugging

internal/controller/etcdbackupschedule_controller.go
```diff
@@ -0,0 +1,263 @@
...
+			ObservedGeneration: schedule.Generation,
+		})
+		return r.updateStatus(ctx, schedule)
+	}
+
+	cronJobName := factory.GetBackupCronJobName(schedule)
+	existingCronJob := &batchv1.CronJob{}
+	err := r.Get(ctx, types.NamespacedName{Name: cronJobName, Namespace: schedule.Namespace}, existingCronJob)
+	if err == nil {
+		return r.reconcileExistingCronJob(ctx, schedule, cluster, existingCronJob)
```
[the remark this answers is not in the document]
BROngineer: It seems it's intended to take over control on the CronJob which matches the naming pattern, but is not owned by the controller.

BROngineer (approved):

api/v1alpha1/etcdbackup_types.go
```diff
@@ -0,0 +1,122 @@
...
+// EtcdBackupStatus defines the observed state of EtcdBackup
+type EtcdBackupStatus struct {
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Cluster",type=string,JSONPath=`.spec.clusterRef.name`
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Complete")].status`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```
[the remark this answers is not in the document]
lllamnyp: This is a valid concern. Maybe surface a top-level `.status.phase` field that stores the current condition, rather that parsing all of the most recent conditions?

api/v1alpha1/etcdbackup_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// SubPath is an optional sub-directory within the PVC volume.
+	// The operator appends the backup filename automatically.
+	// +optional
+	SubPath string `json:"subPath,omitempty"`
+}
+
+// EtcdBackupStatus defines the observed state of EtcdBackup
+type EtcdBackupStatus struct {
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
```
lllamnyp, line 93: Expose an extra field here to use in the printcolumns to cleanly expose the current status.
```suggestion
type EtcdBackupStatus struct {
	Conditions []metav1.Condition    `json:"conditions,omitempty"`
	Phase      EtcdBackupStatusPhase `json:"phase,omitempty"`
}
```

api/v1alpha1/etcdbackup_types.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	EtcdBackupConditionStarted  = "Started"
+	EtcdBackupConditionComplete = "Complete"
+	EtcdBackupConditionFailed   = "Failed"
+)
```
lllamnyp, line 37: Make these conditions a separate type with an underlying string type:
```suggestion
type EtcdBackupStatusPhase string

const (
	EtcdBackupConditionStarted  EtcdBackupStatusPhase = "Started"
	EtcdBackupConditionComplete EtcdBackupStatusPhase = "Complete"
	EtcdBackupConditionFailed   EtcdBackupStatusPhase = "Failed"
)
```

api/v1alpha1/etcdbackup_webhook.go
```diff
@@ -0,0 +1,175 @@
...
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *EtcdBackup) ValidateCreate() (admission.Warnings, error) {
+	etcdbackuplog.Info("validate create", "name", r.Name)
+
+	var allErrors field.ErrorList
+
+	// Job name = "{name}-backup" (7 char suffix).
+	// Job names must be <= 63 chars (DNS label limit).
+	const jobSuffix = "-backup"
+	const maxJobNameLen = 63
```
lllamnyp: This seems to be part of the API. I'd make these constants package-level exported constants and would probably define them in etcdbackup_types.go.

api/v1alpha1/etcdbackupschedule_types.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	EtcdBackupScheduleConditionReady  = "Ready"
+	EtcdBackupScheduleConditionFailed = "Failed"
+)
```
lllamnyp, line 27: Same typing consideration as with the EtcdBackup.

lllamnyp (commented):
**Suggestion: consider using `generateName` instead of deterministic child resource names.**

The current pattern builds a fixed name (`{backup.Name}-backup`, `{schedule.Name}-scheduled-backup`) and looks up the child with a direct `Get()`. This works, but it couples the parent and child names, requires name-length validation in the webhook, and introduces a `JobNameConflict` code path for the edge case where a same-named Job exists but isn't owned by the EtcdBackup.

Since ownership is already fully wired (`ctrl.SetControllerReference` in the factories + `Owns()` in both controllers), switching to `generateName` would be straightforward:

1. **Factory**: use `GenerateName: backup.Name + "-backup-"` instead of `Name`, and add a label like `etcd.aenix.io/etcdbackup-name: <backup.Name>`.
2. **Controllers**: replace the `Get`-by-name with `List` + label selector (or filter by ownerReference).
3. **Drop** `GetBackupJobName` / `GetBackupCronJobName`, the `JobNameConflict` check, and the name-length webhook validation.

There's no idempotency risk: controller-runtime never reconciles the same object concurrently, and the informer cache is updated immediately after `Create`, so a subsequent reconcile's `List` will always find the existing child.

This is a non-blocking suggestion — the current approach works correctly. `generateName` would just remove the naming coupling and simplify the code.

charts/etcd-operator/crds/etcd-backup-schedule.yaml
lllamnyp, line 1: Adding a `.gitattributes` file somewhere nearby with content like
```
charts/etcd-operator/crds/*.yaml linguist-generated
```
would be nice.

charts/etcd-operator/crds/etcd-backup-schedule.yaml
```diff
@@ -0,0 +1,227 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: etcd-operator-system/etcd-operator-serving-cert
+    controller-gen.kubebuilder.io/version: v0.15.0
```
lllamnyp, line 6: TODO: update controller-gen version.

internal/controller/factory/backup_job.go
lllamnyp: I second this. ephemeral-storage would be useful here.

internal/controller/factory/backup_job.go
```diff
@@ -0,0 +1,250 @@
...
+	envVars = append(envVars, corev1.EnvVar{Name: "BACKUP_INCLUDE_REVISION", Value: stringTrue})
+
+	container := corev1.Container{
+		Name:         "backup-agent",
+		Image:        operatorImage,
+		Command:      []string{"/backup-agent"},
+		Env:          envVars,
+		VolumeMounts: volumeMounts,
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
```
lllamnyp, line 249: More specifically, ephemeral-storage requests/limits should be added here.

lllamnyp (changes requested):

charts/etcd-operator/crds/etcd-backup-schedule.yaml
BROngineer: I'd upgrade the whole stack including go itself

lllamnyp: Yeah, but not in this PR, the scope is already quite large.

api/v1alpha1/etcdbackup_types.go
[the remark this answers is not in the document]
BROngineer: ~~I assume that since we'll add the phase field, it will make sense to reflect backup processing steps in conditions:
`EtcdBackupClusterFound`
`EtcdBackupJobSetup`
`EtcdBackupConfig`
etc, rather than backup phase.~~

charts/etcd-operator/crds/etcd-backup-schedule.yaml
BROngineer: Resolving for now, corresponding issue created: https://github.com/aenix-io/etcd-operator/issues/310

api/v1alpha1/etcdbackup_types.go
BROngineer: Fixed

api/v1alpha1/etcdbackup_types.go
BROngineer: Fixed

api/v1alpha1/etcdbackup_types.go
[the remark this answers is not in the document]
BROngineer: Fixed

api/v1alpha1/etcdbackup_webhook.go
BROngineer: Fixed

api/v1alpha1/etcdbackupschedule_types.go
BROngineer: ConditionType is a string. Do we really want to introduce dedicated type and then cast it to string?

internal/controller/factory/backup_job.go
BROngineer: Fixed

lllamnyp (approved):

BROngineer:
Was merged into main ([dcaffd424e5c0a746d9a822f90302accfd016638](https://github.com/aenix-io/etcd-operator/commit/dcaffd424e5c0a746d9a822f90302accfd016638)), but due to gh glitch it's not displayed as merged.

PR #317, opened 2026-05-20
fix(backup): restore-datadir was broken

androndo:
Distroless container in restore-datadir has crashloop, reduce it and added test

## Summary by CodeRabbit

* **New Features**
  * Restore now performs snapshot download and restore in-process via a single init container; init container mounts and env vars adjusted and pod FSGroup set to allow writing to PVCs.

* **Chores**
  * Go toolchain bumped to 1.23.0 across CI, Docker, and module settings; dependency refresh for etcd, zap, OpenTelemetry and related libraries.
  * Restore init-container memory limits increased; linter tool version pinned.

* **Tests**
  * Added unit tests for restore control flow and an e2e test for PVC-based cluster restore.

[Review Change Stack](https://app.coderabbit.ai/change-stack/aenix-io/etcd-operator/pull/317?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cmd/restore-agent/main.go
```diff
@@ -55,19 +58,30 @@ func run() error {
 
 	memberDir := filepath.Join(dataDir, "member")
 	if _, err := os.Stat(memberDir); err == nil {
-		fmt.Println("data directory already exists, skipping restore download")
+		fmt.Println("data directory already exists, skipping restore")
 		return nil
 	}
```
lllamnyp: The skip-check here returns success when `dataDir/member` exists, but `etcdutl`'s `snapshot.Restore()` creates that directory tree very early — before the snapshot bytes are copied, before the SHA-256 check, and before the WAL is written. Concretely, in `go.etcd.io/etcd/etcdutl/v3@v3.5.21/snapshot/v3_snapshot.go`: line 261 sets `s.snapDir = filepath.Join(dataDir, "member", "snap")`, line 274 calls `saveDB()` → `copyAndVerifyDB()`, line 414 does `fileutil.CreateDirAll(s.lg, s.snapDir)` (so `member/snap/` is on disk), line 420 opens `member/snap/db`, line 426 does the big `io.Copy`, lines 442-459 do the SHA-256 check, and only then does line 284 do `saveWALAndSnap()`.

Any midway failure — OOMKill during `io.Copy` (the new 2Gi limit reduces this but doesn't eliminate it for larger clusters), ENOSPC, SHA mismatch, pod eviction, SIGKILL — leaves `member/` on disk in a partial state. On the next init-container attempt, `os.Stat(memberDir)` succeeds, this function returns nil, the main etcd container then boots against the partial data dir, fails to open the backend, and the pod ends up in `CrashLoopBackOff` indefinitely. For PVC-backed storage this state survives pod deletion and recreation and needs manual operator intervention.

The wipe block added in `restoreDataDir()` at lines 279-284 is intended to handle exactly this — and its comment claims *"the skip-check in run() (memberDir exists -> return) protects the post-success case, so this branch only runs on partial failure"* — but the premise is inverted: `member/` exists on *both* a successful prior restore *and* on the partial-failure modes that actually occur, so the run()-level skip fires first and the wipe is unreachable for the cases it was added to handle. `TestRestoreDataDir_WipesPartialDataDir` only exercises stray files in `dataDir` without a `member/` subdir, which is not the state `etcdutl` actually produces on a midway crash.

Two viable fixes:

1. Write a positive sentinel (e.g. `dataDir/.restore-complete`) only after `mgr.Restore()` returns nil, and have the skip-check key off the sentinel rather than `member/`. Cleanly distinguishes completed restore from partial state.
2. Drop the run()-level skip entirely and let `restoreDataDir()`'s wipe-then-restore be idempotent — the wipe path then becomes the single source of truth and the comment at lines 274-278 becomes accurate.

test/e2e/e2e_test.go
```diff
@@ -378,6 +378,106 @@ spec:
...
+			})
+
+			By("wait for source PVCs to drain", func() {
+				Eventually(func() int {
+					cmd := exec.Command("kubectl", "-n", namespace, "get", "pvc",
+						"-l", "app.kubernetes.io/instance=test", "--no-headers")
+					out, _ := utils.Run(cmd)
+					return len(strings.TrimSpace(string(out)))
+				}, 3*time.Minute, 5*time.Second).Should(Equal(0))
+			})
```
lllamnyp: This wait selects on `app.kubernetes.io/instance=test`, but the source `EtcdCluster` is deployed from `examples/manifests/etcdcluster-emptydir.yaml` which uses `storage.emptyDir`. In `internal/controller/factory/statefulset.go`, `CreateOrUpdateStatefulSet` only appends a `volumeClaimTemplate` when `cluster.Spec.Storage.EmptyDir == nil`, so no PVC with that label is ever created. The other PVC in the namespace, `etcd-backup-pvc`, is created from inline YAML at lines 291-302 with no labels at all. As a result `kubectl get pvc -l app.kubernetes.io/instance=test` returns empty on the very first poll, `Should(Equal(0))` matches immediately, and the `Eventually` completes in milliseconds with no actual synchronization.

The spec still works today because the preceding `kubectl delete etcdcluster test --wait=true --timeout=3m` already blocks on cluster teardown and emptyDir state is pod-scoped — there is no persistent state to drain. But this step is dead code that *looks* meaningful. Two footguns:

- If anyone changes the source cluster to persistent storage (a reasonable thing to do — restore-from-PVC is a more interesting test path when the source had real persistent state), the label selector still needs to match whatever the operator labels its `volumeClaimTemplate` with; if conventions drift, the wait silently provides no synchronization while looking like it does.
- A future contributor reading "wait for source PVCs to drain" reasonably assumes the wait is real.

Suggest dropping the step (`--wait=true` on the delete is already sufficient for emptyDir), or switching to a pod-count gate that's meaningful for both storage modes:

```go
By("wait for source pods to drain", func() {
    Eventually(func() int {
        cmd := exec.Command("kubectl", "-n", namespace, "get", "pod",
            "-l", "app.kubernetes.io/instance=test", "--no-headers")
        out, _ := utils.Run(cmd)
        return len(strings.TrimSpace(string(out)))
    }, 3*time.Minute, 5*time.Second).Should(Equal(0))
})
```

Pods carry `app.kubernetes.io/instance` unconditionally via the operator's `PodLabels`, so this gate is meaningful for both emptyDir and PVC-backed clusters.

lllamnyp (changes requested):
Two findings from a deeper read of the restore path — one correctness, one nit. Inline below.

lllamnyp (changes requested):
Reviewed `fix/restore-no-shell` against target branch `main` (merge-base `8f6f1bd`). The branch bundles several backup/restore features: the headless restore-agent fix (the title), `status.snapshot` population from agent pod logs, `ErrInvalidSpec` terminal-failure handling, PVC subPath hardening, FSGroup for restore, OPERATOR_IMAGE kustomize propagation, and a Go 1.23 bump.

The logic is sound and well-tested (all runnable unit tests pass; `go build` and `go vet` are clean; `go mod tidy` is a no-op). But there is a hard blocker: the committed generated artifacts do not match what the project's codegen tooling produces, which fails the CI generate-check.

## Blocking issues

### 1. Generated CRD + deepcopy + docs are stale relative to the source (CI generate-check will fail)

Running the project's pinned codegen (`make manifests generate` → controller-gen v0.15.0, then `make helm-crd-copy`, then `make generate-docs`) mutates four committed files. controller-gen is version-pinned, so this drift is deterministic and not an environment artifact.

- `api/v1alpha1/zz_generated.deepcopy.go` — the `BackupSnapshot.DeepCopyInto`/`DeepCopy` funcs are placed after `EtcdBackupStatus` in the committed file, but controller-gen emits them alphabetically (right after `BackupDestination`, before `BootstrapSpec`). The funcs were hand-added in the wrong spot rather than regenerated. Functionally identical, but the file does not match generator output.
- `config/crd/bases/etcd.aenix.io_etcdbackups.yaml` — the committed CRD's `status.snapshot.uri` description is MISSING the entire trailing paragraph that the current Go doc comment on `BackupSnapshot.URI` contains:
  > "KEEP IN SYNC with the SnapshotURIPrefixPattern constant in this package — the controller compiles its defense-in-depth regex from that constant; if this marker drifts, the controller will admit URIs that the apiserver refuses to write and reconciles will spin on Status().Update validation errors."

  The committed CRD also differs in blank-line handling between description paragraphs. The CRD was generated from an older version of the Go comment, then the comment was edited and `make manifests` was never re-run.
- `charts/etcd-operator/crds/etcd-backup.yaml` — same staleness; it is a copy synced from the config CRD via `make helm-crd-copy` and is missing the same paragraph.
- `site/content/en/docs/v0.4/reference/api.md` — the BackupSnapshot field descriptions are missing the `<br /><br />` paragraph separators (and the "KEEP IN SYNC" sentence) that the current Go comments produce. (Note: a separate `Optional: {}` rendering delta also appears locally because the Makefile pins `crd-ref-docs@latest` unpinned — that part is an environment artifact and NOT PR-owned. The paragraph-break / KEEP-IN-SYNC delta IS PR-owned because it derives from the source comment, independent of the crd-ref-docs version.)

Fix: regenerate, do not hand-edit. Run `make manifests generate helm-crd-copy generate-docs` and commit the result. Verify with a clean tree afterward. This is exactly what the repo's existing generate-drift CI check catches (`make manifests generate` followed by `git diff --exit-code`) — no new test is needed, the contract is already guarded at codegen time.

## Non-blocking observations (recommended)

- `internal/controller/etcdbackup_controller.go:415` — the `errNoMarker` grace branch requires `job.Status.CompletionTime != nil`; a nil CompletionTime falls straight through to finalize-with-empty, skipping the grace window. In practice the Job controller always sets `CompletionTime` alongside `Succeeded>=1`, so this is unreachable, but if you want belt-and-suspenders, treat nil CompletionTime as "still within grace" (requeue) rather than "finalize empty".
- Restore-side `PVC.SubPath` (`internal/controller/factory/statefulset.go:540-548`) is interpolated into `PVC_BACKUP_PATH` without the `validatePVCSubPath` guard that backup destinations now get. The restore CR is operator-authored and the value only selects a file within the same already-mounted PVC, so the blast radius is small — but applying the same validation symmetrically would be consistent with the hardening this PR introduces.

## Things verified as correct

- restore-agent `.restore-complete` sentinel logic: skip-on-sentinel in `run()`, wipe-stray-dir + `snapshot.Restore()` + sentinel-write in `restoreDataDir()`. The sentinel lands at `/var/run/etcd/default.etcd/.restore-complete`, alongside `member/`; etcd ignores extra top-level files. etcd container `--data-dir` and restore-agent `ETCD_DATA_DIR` agree (`/var/run/etcd/default.etcd`).
- `snapshot.RestoreConfig` field usage matches the vendored etcdutl v3.5.21 API; the doc explicitly states `OutputDataDir` must not pre-exist, which is exactly why the wipe is required.
- `ETCD_INITIAL_CLUSTER` / `ETCD_INITIAL_CLUSTER_TOKEN` are present in the cluster-state ConfigMap (`configmap.go:92-93`) and reach the agent via `EnvFrom`; `HEADLESS_SVC`, `POD_NAME`, `POD_NAMESPACE` are wired in `generateRestoreInitContainers`.
- Marker round-trip: agent emits `%q`-quoted URI; controller regex captures the quoted form, `strconv.Unquote`s it, and applies the shared `SnapshotURIPrefixPattern` as defense-in-depth. Spaces/escaped quotes round-trip (unit-tested).
- `reconcileJobStatus` sentinel branching (errPodGone/errPodNotReady/errNoMarker/transient) is correct; `Reconcile` early-returns on Phase==Complete||Failed so finalized backups are not re-extracted.
- OPERATOR_IMAGE: Helm chart sets it in `deployment.yml`; kustomize `replacements` keeps `config/manager/manager.yaml` in sync with the manager image. RBAC for `pods` get;list and `pods/log` get added in both `config/rbac/role.yaml` and the Helm clusterrole.
- kamaji example: moving `127.0.0.1` from `dnsNames` to `ipAddresses` is correct for the cert-manager Certificate spec.
- Go 1.23.0 bump is consistent across CI workflows, Dockerfile, and `go.mod`.

## Could not verify in this environment

The envtest-based controller suite (`internal/controller`, 31 specs including the new status.snapshot extraction integration tests and CRD round-trip validation tests) could not run here: `fork/exec` of the envtest control-plane binaries is blocked locally (`bin/k8s/1.30.0-linux-amd64/etcd` exists and is executable, but exec is denied). The CRD round-trip tests (`Should agree between the controller URI regex and the CRD URI pattern`, `Should reject an empty URI`, `Should reject a too-short checksum`, `Should accept hyphenated checksum algorithms`) validate the installed CRD schema — these must be run in CI, and they will only pass once the CRD is regenerated (blocker 1), since they exercise the apiserver-installed schema.

lllamnyp (approved):
Both findings from my earlier change-request are resolved in `3f447d1`: the restore skip-check now keys off a `.restore-complete` sentinel written only after `snapshot.Restore()` succeeds (so a mid-restore crash leaves no sentinel and the wipe-then-retry path is reachable), and the e2e drain wait switched from the never-matching PVC selector to a pod selector that is meaningful for both emptyDir and PVC storage. LGTM.

## Scope

This branch carries more than the `fix/restore-no-shell` name implies. The 42-file commit `3f447d1` bundles three things:
1. The titular fix: collapse the two-initContainer restore flow (download via `restore-agent` + `etcdutl snapshot restore` via a `/bin/sh -c` initContainer on the distroless etcd image, which has no `/bin/sh` and crash-looped) into a single `restore-agent` that downloads and runs `etcdutl`'s `snapshot.Restore()` Go API in-process.
2. A new `EtcdBackupStatus.snapshot` observational sub-resource (URI / sizeBytes / checksum), populated by parsing the backup-agent's terminal pod-log marker.
3. Toolchain bumps (Go 1.23 in Dockerfile, controller-tools v0.16.5, golangci-lint v1.60.0) and a new `codegen-drift.yaml` CI gate.

Mixing a focused fix with a sizeable feature in one commit is a packaging nit, not a blocker. The work itself is correct and unusually well-tested.

## What I verified in detail (no blockers found)

**Restore peer-URL / initial-cluster consistency** (`cmd/restore-agent/main.go`, `internal/controller/factory/configmap.go:72-79`): the agent builds `peerURL = https://<podName>.<headlessSvc>.<podNamespace>.svc:2380` and passes it as `PeerURLs` to `snapshot.Restore`. etcdutl requires this to match the member's entry in `--initial-cluster`. The cluster-state ConfigMap builds each member as `https://<pod>.<headlessSvc>.<namespace>.svc:2380` — identical shape, matching the legacy script's `--initial-advertise-peer-urls`, so behavior is preserved.

**Sentinel-vs-member skip logic** (restore-agent `run()` and `restoreDataDir()`): the skip-check correctly keys off `.restore-complete` rather than `member/`, because etcdutl creates `member/snap/` before the SHA-256 check and WAL write — a mid-restore crash leaves a partial `member/` that the old `[ -d member ]` test would have treated as "done", permanently wedging retries. The sentinel is written only after `Restore()` returns nil; earlier failure leaves no sentinel and the `os.RemoveAll(dataDir)` wipe reclaims the dir. The wipe targets only the dataDir; the downloaded snapshot lives in a separate `/restore` emptyDir, so it cannot be destroyed. The stray `.restore-complete` dotfile is harmless to etcd (it looks for `member/`, ignores sibling files). Covered by the restore-agent tests.

**FSGroup gating** (`statefulset.go`): `FSGroup=65532` is set only when `bootstrap.restore != nil`. Locked in both directions by `statefulset_restore_test.go`.

**Marker parser contract** (`etcdbackup_controller.go`; agent emitters in `backup-agent/main.go`): the controller regex matches exactly the agent's `%q`-quoted emit strings; URI is unquoted so keys with spaces/quotes round-trip; the right anchor rejects over-length sha256; `+` rejects `uri=""`. A defense-in-depth regex compiled from the shared `SnapshotURIPrefixPattern` constant drops non-`s3://`/`file://` URIs before they could fail `Status().Update`. Exhaustively tested in `etcdbackup_log_scan_test.go`.

**Reconcile idempotency & failure classification** (`etcdbackup_controller.go`): Complete/Failed phases return early, so the log scan never repeats. The `errPodNotReady`/`errPodGone`/`errNoMarker` branching (requeue / finalize-empty / backoff) is sound, including the `nil CompletionTime` belt-and-suspenders requeue. `ImagePullPolicy: IfNotPresent` pins the agent to the manager's image bytes.

**RBAC / manifests / docs**: no codegen drift (`make manifests generate helm-crd-copy` is clean); both `config/rbac/role.yaml` and the chart clusterrole carry the new `pods`/`pods/log` grants (no `watch`, intentionally). Generated `api.md` reflects the new types; the `type.tpl` change correctly escapes the `|` in the `^(s3|file)://.+` pattern (which would otherwise break the markdown table) and filters dangling anchors. `kustomization.yaml` + `manager.yaml` keep `OPERATOR_IMAGE` in sync with the manager image. `etcdcluster-kamaji.yaml` correctly moves `127.0.0.1` to cert-manager `ipAddresses`.

**End-to-end coverage** (`test/e2e/e2e_test.go`): a new ordered Context runs full PVC backup → assert `status.snapshot.*` → delete source cluster → wait for source pods to drain → re-create with `bootstrap.restore` → fast-fail on initContainer CrashLoopBackOff → wait Ready. This is the correct place to guard the no-shell fix and the `backup-agent` container-name wire contract.

## Recommended (non-blocking)

- Commit packaging: the toolchain bumps and the kamaji example fix are unrelated to the restore fix / snapshot-status feature; splitting them out would make bisect/changelog cleaner.
- The `BackupSnapshot` field doc comments are very long (rendered API table cells are paragraphs) — accurate, just dense.

PR #318, opened 2026-05-26
chore: migration tracker — relocate etcd-operator to cozystack (etcd-operator.cozystack.io/v1alpha2)

kvaps:
> **Tracking PR — do not merge until the checklist is complete.**
> This is a living checklist coordinating the migration of `aenix-io/etcd-operator`
> onto the new code base derived from `lllamnyp/etcd-operator` and its relocation
> into the `cozystack` GitHub organization.
>
> New API: group `etcd-operator.cozystack.io`, version `v1alpha2`.
> Each item below is tracked by a dedicated issue/PR linked inline as work starts.

## 1. Feature-parity analysis
- [x] Diff the feature sets of `aenix-io/etcd-operator` and `lllamnyp/etcd-operator`
- [x] Confirm features present in aenix but missing in lllamnyp: **multi-user**, **backups**
- [x] ~~Support the **multi-user for Kamaji** use case (per-tenant etcd credentials)~~ cover kamaji use case with integration test
- [x] Exclude `*Template` fields from the parity scope (intentionally not needed)

> Parity analysis posted as a comment below. Integration branch: `v1`
> (imported from `lllamnyp/etcd-operator` with its history intact);
> feature PRs target `v1`.

## 2. New fields to add
- [x] `additionalMetadata`
- [x] `affinity`
- [x] `topologySpreadConstraints`
- [x] additional configuration options

### cozystack used
  - [x] options:
    quota-backend-bytes: {{ include "calculateQuotaBackendBytes" .Values.size | quote }}
    auto-compaction-mode: "periodic"
    auto-compaction-retention: "5m"
    snapshot-count: "10000"

## 3. Backup CronJob ownership
- [x] Keep the backup CronJob managed by Cozystack (not by the operator)
- [ ] Open an issue describing this decision
- [ ] Add it to the roadmap

## 4. Codebase convergence
- [x] Port `aenix-io/etcd-operator` to the `lllamnyp/etcd-operator` code base → imported into branch `v1` (history intact)
- [x] Rename API group to `etcd-operator.cozystack.io` → #319 (open)
- [x] Bump API version to `v1alpha2` → already `v1alpha2` (carried from lllamnyp)
- [x] kubectl etcd plugin

## 5. Migration / upgrade path
- [x] Choose a strategy (conversion webhook vs. alternatives)
- [x] Implement conversion `etcd.aenix.io/v1alpha1` → `etcd-operator.cozystack.io/v1alpha2`
- [x] Document the upgrade procedure for existing users
[...]

PR #319, opened 2026-05-26
chore(api): rename group to etcd-operator.cozystack.io and module to cozystack

kvaps:
## What

Mechanical rename of the freshly imported `v1` base (originally `lllamnyp/etcd-operator`) onto the Cozystack identity. No behavioural changes.

- **API group:** `lllamnyp.su` → `etcd-operator.cozystack.io` (version `v1alpha2` unchanged)
- **Go module:** `github.com/lllamnyp/etcd-operator` → `github.com/cozystack/etcd-operator`
- **Label / finalizer keys:** now prefixed with the API group itself — `etcd-operator.cozystack.io/cluster`, `/role`, `/member-cleanup` — instead of the previous `etcd.lllamnyp.su/…` form. **Decision open to review:** dropped the redundant `etcd.` segment rather than producing `etcd.etcd-operator.cozystack.io/…`.
- CRD manifests, RBAC, samples and deepcopy regenerated (`make manifests generate`).
- Upstream issue/PR cross-links in `docs/` and `README.md` deliberately left pointing at `lllamnyp/etcd-operator` (those issues live there; docs get rewritten in a later phase).

## Out of scope

- Docker image name in `.github/workflows/publish.yml` (still `lllamnyp/etcd-operator`) — CI/publishing rename is a separate change.
- The go module path `github.com/cozystack/etcd-operator` will not resolve via `go get` until `cozystack/etcd-operator` exists; local builds and CI are unaffected.

## Verification

- `go build ./...` ✓
- `go vet ./...` ✓
- `make test` (envtest, k8s 1.33) ✓ — all packages green

Part of the migration tracked in #318. Targets the `v1` integration branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,9 +139,9 @@ KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 
-# go-install-tool installs $2@$3 under $1. Pattern stolen from
-# lllamnyp/operator. `go install` drops the binary at $LOCALBIN/<basename>,
-# so we rename it after install to the version-suffixed target path.
+# go-install-tool installs $2@$3 under $1. `go install` drops the binary at
+# $LOCALBIN/<basename>, so we rename it after install to the version-suffixed
+# target path.
 define go-install-tool
 @[ -f $(1) ] || { \
 set -e; \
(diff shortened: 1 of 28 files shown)


lllamnyp:
A couple of old-identity leftovers survive the rename:

- `main.go:40` / `main.go:55` — the import path is correctly renamed to `github.com/cozystack/etcd-operator/api/v1alpha2`, but the alias is still `lllamnypsuv1alpha2`. Renaming it (e.g. `etcdv1alpha2`) keeps the old org name out of the renamed tree.
- `Makefile:143` — a comment still references `lllamnyp/operator`.

Both are harmless to builds; flagging in case you'd rather fold them into this mechanical rename than leave the old name behind.

On the open question in the description: dropping the redundant `etcd.` segment reads fine — the group already carries `etcd-operator`, so `etcd.etcd-operator.cozystack.io/…` would have been the awkward alternative. The `etcd-operator.cozystack.io/{cluster,role,member-cleanup}` keys land consistently across the label/finalizer constants, the `/scale` `status.selector`, the RBAC markers, the CRD filenames, and the samples.

lllamnyp (approved):

PR #320, opened 2026-06-01
feat(auth): support option to enable single-user authentication

androndo:
Support spec.security.enableAuth with referencing to secret on the root creds


api/v1alpha2/cel_validation_test.go
```diff
@@ -573,3 +573,143 @@ func TestCEL_HappyPathAccepts(t *testing.T) {
...
+		},
+	}
+	c.Spec.Security = &lll.SecuritySpec{EnableAuth: true} // no rootCredentialsSecretRef
+
+	err := k8s.Create(ctx, c)
+	if err == nil {
+		_ = k8s.Delete(ctx, c)
+		t.Fatalf("apiserver accepted enableAuth without rootCredentialsSecretRef; expected rejection")
+	}
+	if !strings.Contains(err.Error(), "requires spec.security.rootCredentialsSecretRef") {
```
lllamnyp: Don't hardcode magic strings like that. If they're present here, they're also present in the named error.

lllamnyp (changes requested):
The single-user root-auth implementation is solid — the `status.authEnabled` latch cleanly separates intent (`spec.*.enableAuth`) from the runtime dial decision, the bootstrap-window anonymous dial is correct, and the crash-recovery path through `AuthStatus` + `isAuthRequiredErr` is well-reasoned and well-tested. The client-side provisioning sequence (`UserAdd("root", …)` → `UserGrantRole("root", "root")` → `AuthEnable`) maps exactly onto etcd's own client API, and relying on the built-in `root` role to skip `RoleAdd` is the right call. One blocker before merge, plus one real-but-out-of-scope item.

## Blocking: name the API subtree `auth`, not `security`

etcd has no concept called "security." The entire domain — users, roles, `auth enable`, `auth status` — is "auth"/"authentication" in etcd's CLI, client API, and docs. The CRD should read the way an etcd operator expects:

- `spec.security` → `spec.auth`; `SecuritySpec` → `AuthSpec`
- `spec.security.enableAuth` → `spec.auth.enabled` — this mirrors `etcdctl auth enable` and the `AuthStatusResponse.Enabled` field the code already reads, and drops the awkward `security.enableAuth` verb doubling
- `status.authEnabled` already matches `AuthStatus().Enabled` — keep it
- `rootCredentialsSecretRef` already matches etcd's built-in `root` user — keep it

This is the moment to do it: nothing has shipped on this API yet, so the rename is free — no stored-object migration, no deprecation window. The change touches the two type fields, the four CEL `XValidation` rules, the regenerated deepcopy + CRD (`make manifests generate`, not hand-edited), and the docs/tests. After it lands, the field name and the etcd command it drives will read identically.

## Out of scope (please file a follow-up, don't fix here)

The branch still carries stale old-identity references that the rename commits missed — `lllamnyp/etcd-operator` issue links in `README.md`, `docs/concepts.md`, `docs/installation.md`, `docs/operations.md`, and a `.github/workflows/publish.yml` that builds/pushes to the `lllamnyp/etcd-operator` Docker Hub repo on a `master` trigger that no longer exists. These are real (the docs links now contradict the `cozystack/etcd-operator` links the Go source was updated to, and the publish path would push to a repo the project doesn't own), but they're leftovers from the identity rename, not part of the auth feature. They belong in a dedicated cleanup PR that can reconcile *all* references at once — including deciding the canonical home (`cozystack/etcd-operator` vs the `aenix-io/etcd-operator` git remote) so docs and source agree. Folding that into this PR would muddy an otherwise clean feature diff. Could you open a follow-up issue/PR to scrub them properly?

lllamnyp (approved):
Blocker resolved. The `spec.security` → `spec.auth` rename is complete and consistent with upstream etcd: `AuthSpec`, `spec.auth.enabled` (mirroring `etcdctl auth enable` / `AuthStatusResponse.Enabled`), all four CEL `XValidation` rules, the controller/client field references, and the regenerated deepcopy + CRD. `make manifests generate` produces no drift and `go build ./...` passes. Nice touch extracting the CEL rejection messages into the `MsgAuth*` constants — the envtest CEL tests still assert against the real apiserver rejection text, so it's a literal-dedup, not a source-parsing guard.

The old-identity (`lllamnyp/etcd-operator`) leftovers in the docs and `publish.yml` were correctly left untouched for the separate cleanup PR. Please open that follow-up so the issue links and the Docker Hub publish target get reconciled to the canonical repo across the board.

LGTM.

PR #321, opened 2026-06-02
feat(snapshots): add snapshot and restore support, ported from the previous implementation

androndo:


lllamnyp (changes requested):
The implementation is in good shape and I'm confident in the mechanics — one blocking request, which is about naming, not correctness.

## What's solid

- **Restore idempotency / data safety**: `RunRestore` no-ops when `member/` already exists (survives Pod restarts), stages to `.restore` and renames `member/` into place, cleans stale crash debris before the free-space pre-flight, and rejects a negative `HeadObject` ContentLength before the `uint64` wrap could spuriously pass the space check. Memory-medium + restore is CEL-rejected. Well covered by `restore_test.go` and `backup_cel_test.go`.
- **Overwrite guards**: both S3 (`ensureObjectAbsent` via HeadObject + UID object-metadata) and PVC (`ensureFileAbsent` + `.uid` sidecar) refuse to clobber another snapshot while treating a same-UID retry as idempotent, and the S3 guard fails closed on a non-404. Atomic PVC write (temp + rename + fsync).
- **Hang bounding**: agent `BackupTimeout` (25m) under the Job `ActiveDeadlineSeconds` (30m); restore `RestoreTimeout` (30m) as the sole guard where there is no Job backstop — correctly reasoned.
- **Auth-snapshot restore**: the discovery-error branches distinguish "auth required, none configured" (terminal) from "credentials rejected" (requeues), with the operator-facing messages pointing at the real misconfiguration (the root password must be the snapshot's original, since etcd stores a bcrypt hash).
- **Security & image wiring**: backup Job and seed Pod both run non-root 65532, no SA token, dropped caps, seccomp RuntimeDefault; both controllers fail loudly on an unconfigured operator image. RBAC grants exactly what's used.

## Blocking: rename `Backup` → `Snapshot` to match upstream etcd

Upstream etcd has exactly one term for what this feature does, and it isn't "backup":

- CLI: `etcdctl snapshot save` / `snapshot restore` / `snapshot status`
- client API: `Maintenance.Snapshot(ctx)` — the RPC this agent calls (`cli.Snapshot(ctx)`)
- restore API: the `etcdutl/v3/snapshot` package — `snapshot.NewV3()`, `snapshot.RestoreConfig{SnapshotPath, ...}`, exactly what `RunRestore` uses

`etcdutl` *does* still ship a `backup` command, but its own help is `"[legacy] offline backup of etcd directory"` with `Long: "Prefer: etcdctl snapshot save instead."` — i.e. in upstream, "backup" names the deprecated etcd-v2 offline data-dir copy. This feature is built entirely on the modern `Snapshot` RPC + `snapshot.Restore`, so it implements *snapshot* but is named after the one thing upstream tells users not to use. The code already half-agrees: the agent's success marker prints `snapshot uploaded: ...` and the status sub-record is `BackupSnapshot`. The inconsistency is only at the API/type layer.

There's also a layering reason to free up the word "backup": "snapshot" is the right altitude for *this* operator — a single etcd point-in-time artifact. "Backup" belongs one level up, where a higher-level API composes these snapshots into a backup of a larger system (e.g. a Cozystack Backup that snapshots the etcd it manages via this operator). Naming this layer `Snapshot` keeps "Backup" available for that abstraction instead of colliding with it.

Requested renames (`restore` already matches upstream — `BootstrapSpec.Restore`, `RestoreSpec`, `spec.bootstrap.restore`, `restore-agent` — and should stay):

| Current | Rename to |
|---|---|
| `EtcdBackup` (+ `…Spec`/`…Status`/`…List`) | `EtcdSnapshot` (+ …) |
| `EtcdBackupStatusPhase` (+ phase consts) | `EtcdSnapshotStatusPhase` (+ …) |
| `BackupReady` condition | `SnapshotReady` |
| `BackupDestination` | `SnapshotLocation` — direction-neutral; it's used as both a write `Destination` and a restore `Source`, so a name that asserts neither reads correctly in both roles |
| `S3BackupDestination` / `PVCBackupDestination` | `S3SnapshotLocation` / `PVCSnapshotLocation` |
| `EtcdBackupReconciler`, `buildBackupJob`, `backupJob*`, `RunBackup`, `backup-agent` | `EtcdSnapshot…` / `buildSnapshotJob` / `RunSnapshot` / `snapshot-agent` |

The one spot that needs a real name rather than a substitution: `BackupSnapshot` (the stored-artifact record: `URI`/`SizeBytes`/`Checksum`) lives in `status.snapshot`. Under the kind `EtcdSnapshot`, a `status.snapshot` of type `Snapshot` is "the snapshot's snapshot." Please rename the field to **`status.artifact`** and the type to **`SnapshotArtifact`**.

The field names `Destination` (on the snapshot spec) and `Source` (on the restore spec) are fine to keep — they're the roles; `SnapshotLocation` is the shared type they point at.

This is the moment to do it: nothing has shipped on v1alpha2 yet, so the rename is free — no stored-object migration, no deprecation window. It's a mechanical change verified by the existing suite (everything compiles against the new names, the CEL message-constant tests carry over, and `make manifests generate` must stay drift-free after the CRD kind/RBAC resource rename). The internal controller↔agent env contract (`BACKUP_DEST_KIND`/`BACKUP_NAME`/`BACKUP_UID`) is not user-facing; sweeping it for consistency is welcome but optional.

## Non-blocking

- Commit subject typo: `add backup-resotre` → `restore`. It becomes the squash subject, so worth fixing.

lllamnyp (approved):
Approving. The `Backup` → `Snapshot` rename is complete and consistent with upstream etcd, and I verified the change is a pure mechanical rename with no logic drift (old vs new files are identical after normalizing the renamed identifiers, plus the intended `status.snapshot` → `status.artifact` field rename).

Verified on the current branch:
- Kind `EtcdBackup` → `EtcdSnapshot` (+ `Spec`/`Status`/`List`), phase type + consts, and the `SnapshotReady` condition.
- `BackupDestination` → `SnapshotLocation` (direction-neutral, used as both the snapshot `Destination` and the restore `Source`), with `S3SnapshotLocation` / `PVCSnapshotLocation`.
- Status artifact record `BackupSnapshot` → `SnapshotArtifact`, surfaced at `status.artifact`.
- Controller/job/agent symbols (`EtcdSnapshotReconciler`, `buildSnapshotJob`, `RunSnapshot`, `snapshot-agent`) and the internal env contract (`SNAPSHOT_DEST_KIND`/`NAME`/`UID`).
- `restore` naming kept throughout (`spec.bootstrap.restore.source`), matching `etcdctl snapshot restore`.
- CRD (`…_etcdsnapshots.yaml`), RBAC resource (`etcdsnapshots`), kustomize, deepcopy regenerated — `make manifests generate` is drift-free; `go build` and the api/controllers/internal test suites pass. The migration doc now records the rename for anyone on a pre-rename build.

I went ahead and fixed the two non-blocking nits directly on the branch and force-pushed (the three commits were rebuilt, original authorship preserved):
- The feature commit subject — corrected the `resotre` typo and the now-stale `backups` wording to `feat(snapshots): add snapshot and restore support, ported from the previous implementation`. The PR title is updated to match so the squash subject lands clean.
- The stale code comment in `internal/agent/agent.go` that still pointed at `controllers/backup_job.go` now reads `controllers/snapshot_job.go`.

Setting auto-merge (squash). LGTM.

PR #322, opened 2026-06-03
fix(main): add extra labels/annotations for created object and kubectl plugin

androndo:


lllamnyp (changes requested):
The two features are well-built — the operator-label shadowing guard is correct and tested, affinity/TSC latch cleanly through `ObservedClusterSpec`, and the plugin's port-forward and RBAC-failure handling are sound. Five things to address before merge.

## Blocking

### 1. `additionalMetadata` is not applied to PVCs

The field doc says the metadata is merged onto "every object it creates for this cluster," but the per-member PVC (`etcdmember_controller.go:371`) is built with `Labels: memberLabels(...)` only — no `applyAdditionalMetadata`, and no annotations at all. PVCs are operator-created and operator-owned (`SetControllerReference` + `Owns(&PersistentVolumeClaim{})`), and they're a prime target for this feature (backup-tool selectors, cost-allocation labels). Apply the metadata in the PVC build the same way the Pod does (`etcdmember_controller.go:628`), reading from `member.Spec.AdditionalMetadata`. Memory-medium members correctly have no PVC, so this only affects PVC-backed members. Add a member-controller test asserting the PVC carries the labels/annotations.

### 2. `additionalMetadata` reads live while its siblings latch — reconcile or document

At member creation (`etcdcluster_controller.go:416` and `808`), `AdditionalMetadata` is read from `cluster.Spec` while `Affinity`, `TopologySpreadConstraints`, `Resources`, `Version`, and `Storage` in the same struct literal are read from `cluster.Status.Observed`. Either route `additionalMetadata` through `ObservedClusterSpec` like the others (consistent, and `specEqualsObserved` / `snapshotSpecIntoObserved` already exist to extend), or, if the live read is deliberate, add a comment at the field and the call sites explaining why metadata intentionally bypasses the target latch. As-is it reads as an oversight.

### 3. Operator annotations should win over `additionalMetadata`, like labels do

`applyAdditionalMetadata` (`helpers.go:241`) protects operator-owned label keys on collision but applies annotations with no precedence. Make it symmetric — take the object's existing annotations and let operator keys win — so user-supplied annotations can never shadow an operator annotation. No operator-owned annotations exist today, so nothing is being clobbered yet; this is preventive, but the label/annotation asymmetry shouldn't ship. Unit-test the collision for both maps.

### 4. The plugin re-implements the operator's etcd client/TLS/auth assembly

`cmd/kubectl-etcd/main.go` carries its own `setupEtcdClient` / `getTLSConfig` / `loadCredentials` / `extractTLSFiles`, duplicating the dial, TLS-config, and credential logic in `controllers/etcd_client.go` (`DefaultEtcdClientFactory`, `buildOperatorTLSConfig`, `resolveEtcdCredentials`). The transport differs legitimately (the plugin port-forwards; the operator dials the Service), but the clientv3 + TLS + auth wiring is the same and will drift — the auth/TLS handling in particular was just built carefully on the operator side. Factor the shared construction into a package both import; let the plugin keep only its port-forward transport.

### 5. New fields ship undocumented, and contradict existing docs

The PR adds `spec.affinity`, `spec.topologySpreadConstraints`, and `spec.additionalMetadata`, but the docs aren't reconciled:

- `additionalMetadata` appears nowhere in `docs/` — only in the type's go-doc and the commented sample. It needs a `concepts.md` entry covering the merge semantics (operator keys win, apply-on-create, which objects are stamped).
- The new scheduling fields contradict current docs that still describe them as unavailable:
  - `operations.md:561` calls a native anti-affinity option "a future feature" and recommends a mutating-webhook workaround — both now false.
  - `operations.md:431` still walks operators through the webhook workaround.
  - `installation.md:123` and the README "What's not supported" section list anti-affinity as a remaining gap.

Update these to point at `spec.affinity` / `spec.topologySpreadConstraints`, and add the `additionalMetadata` documentation. (Leave the `lllamnyp/etcd-operator` issue-link cleanup in those same lines to the separate identity-scrub follow-up — not this PR.)

lllamnyp (changes requested):
The previous round is fully addressed — PVC stamping, the Observed-latch for `additionalMetadata`, symmetric annotation precedence, the shared `internal/etcdclient` package, and the docs reconciliation all look good. One new blocker surfaced in the plugin's TLS handling.

## Blocking: `kubectl-etcd` dials plaintext against a server-TLS-only cluster

`getTLSConfig` → `findSecretNameForTLS` (`cmd/kubectl-etcd/main.go`) decides whether the cluster uses TLS *solely* by looking for a `--trusted-ca-file=` flag on the etcd container. When it's absent, `findSecretNameForTLS` returns `errNoTrustedCAFile` and `getTLSConfig` treats that as "plaintext cluster" — returns `(nil, nil)`, so `etcdclient.New(..., nil, ...)` dials plaintext.

But this operator emits `--trusted-ca-file` **only in mutual-TLS mode** (`etcdmember_controller.go:659-667`):

```go
if clientTLS {
    cmd = append(cmd, "--cert-file=...", "--key-file=...")
    if clientMTLS {                       // only here
        cmd = append(cmd, "--client-cert-auth=true", "--trusted-ca-file=...")
    }
}
```

`clientTLS && !clientMTLS` is the **server-TLS-only** mode the project supports (`clientMTLS := clientTLS && member.Spec.TLS != nil && member.Spec.TLS.ClientMTLS`, `etcdmember_controller.go:597`). In that mode etcd serves HTTPS on 2379 with `--cert-file`/`--key-file` and no `--trusted-ca-file`, so the plugin hits the `errNoTrustedCAFile` branch and dials plaintext at a TLS listener — every subcommand fails the handshake. (The detection was carried over from the previous plugin, whose operator had a different TLS model; it doesn't match this operator's server-TLS-only path.)

**Fix:** detect TLS on a flag present in *both* modes — `--cert-file=` — and resolve the secret from the `tls-client` mount (`/etc/etcd/tls/client`, secret `member.Spec.TLS.ClientServerSecretRef`, `etcdmember_controller.go:670-680`), which carries `ca.crt`/`tls.crt`/`tls.key` in both modes. The client keypair should only be *required* when the server actually demands it (mTLS, i.e. `--client-cert-auth`/`--trusted-ca-file` present); for server-TLS-only, `ca.crt` alone is enough to build a `RootCAs`-only config.

**Test:** unit-test `getTLSConfig` with a synthetic Pod whose etcd container has `--cert-file`/`--key-file` but no `--trusted-ca-file`, plus the `tls-client` volume and a populated secret; assert a non-nil `*tls.Config` with `RootCAs` set (today it returns `(nil, nil)`).

## Non-blocking

- **`forfeit-leadership` can target a learner** (`cmd/kubectl-etcd/main.go`): it transfers to the first non-leader member without filtering `IsLearner`, but `MoveLeader` to a learner is rejected by etcd, so the command fails if it picks a learner mid-scale-up. Skip learners when choosing the target. Worth fixing here while the plugin is in flight; test with a `MemberList` fake of leader + learner + voter asserting the voter is chosen.

lllamnyp (approved):
The last open blocker is resolved. The `kubectl-etcd` TLS detection now keys on `--cert-file` (present in both client-TLS modes) rather than `--trusted-ca-file` (mTLS-only), with a separate mTLS signal: server-TLS-only clusters get a RootCAs-only config from `ca.crt`, mTLS additionally loads the client keypair, and a missing keypair under mTLS fails up front with a clear message. The case matrix is covered by tests — server-TLS-only (non-nil config, RootCAs set, zero client certs), mTLS, mTLS-missing-keypair, and plaintext. The `forfeit-leadership` learner-skip fix from the same review round is included and tested.

Scope is clean — only the two plugin files changed, no controller/API/manifest churn. Build, vet, and tests pass; no codegen drift.

Together with the earlier round (PVC stamping, the Observed-latch for `additionalMetadata`, symmetric annotation precedence, the shared `internal/etcdclient` package, and the docs reconciliation), everything raised on this PR is addressed.

LGTM.

PR #323, opened 2026-06-04
feat(main): add e2e with kamaji case

androndo:
inspired by cozystack use-case with Kamaji DataStore and TenantControlPlane

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,11 @@
 name: CI
 
 on:
+  # No PR branch filter: PRs here target main and integration branches (e.g.
+  # v1) alike; the old `master` filter matched nothing, so CI never ran on PRs.
   pull_request:
-    branches: [ master ]
   push:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   verify:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,48 @@
+name: E2E
+
+on:
+  # No PR branch filter: PRs in this repo target main as well as integration
+  # branches (e.g. v1), and the suite must gate all of them. A filter naming
+  # a branch PRs never target (this said `master`, which does not exist) means
+  # the workflow never runs at all.
+  #
+  # paths-ignore: a kind+cert-manager+Kamaji provisioning run costs ~30-45
+  # minutes; skip it for PRs that touch nothing the suite exercises
+  # (docs-only changes). Everything else — Go code, manifests, the harness,
+  # the workflows themselves — still gates.
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+  push:
+    tags: [ "v*" ]
+  workflow_dispatch:
+
+concurrency:
+  group: e2e-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  kamaji-datastore:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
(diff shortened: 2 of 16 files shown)


lllamnyp (changes requested):
The e2e suite here is excellent and exactly the right shape — more on that below. But this branch was cut from an early point on `fix/align_to_prev` (before that PR's outstanding review fixes), and that creates two ordering problems that need to be resolved before this can be reviewed fairly, plus one change that should be dropped outright.

## Blocking

### 1. Revert the `publish.yml` change

`publish.yml` shouldn't be touched in this PR at all. The commit flips its trigger from `master` to `main` while the image is still `lllamnyp/etcd-operator` and the push authenticates with `secrets.DOCKER_USERNAME` (`publish.yml:18,28`). Today that workflow is harmlessly dead (the `master` filter matches nothing); this change arms it to build and push `latest` to a Docker Hub repo the project no longer owns. The registry/identity cleanup is a separate, tracked concern — please drop the `publish.yml` edit here and leave it to that follow-up. (The `ci.yml` and `e2e.yml` trigger changes are fine to keep — they only run verification, with no outbound side effect.)

### 2. Rebase onto the updated `fix/align_to_prev` first

This branch's merge-base is `b417c71`, but `fix/align_to_prev` has since advanced (and has one more TLS fix still in flight). Two parts of this diff depend on that newer base and can't be assessed fairly against the stale fork point:

- **Plugin TLS tests.** The new `cmd/kubectl-etcd/helpers_test.go` pins the current `--trusted-ca-file`-only TLS detection (`TestFindSecretNameForTLS_NoCAFlag` asserts the `errNoTrustedCAFile` sentinel; `TestGetTLSConfig_*` cover only plaintext and mTLS, with no server-TLS-only case). That detection is being corrected on the base branch to key on `--cert-file` and resolve from the `tls-client` mount, returning a non-nil `RootCAs` config for server-TLS-only clusters. Once that lands, these assertions are wrong and must be updated — and a server-TLS-only case added. The tests aren't incorrect against the code they ship with today; they just encode behavior the base is about to change.
- **Seed `IsVoter` controller change.** Switching the seed's `IsVoter` stamp from `Status().Update` to `Status().Patch(MergeFrom)` is a genuine, correct fix (and nicely covered by `TestBootstrap_SeedIsVoterStampSurvivesConcurrentStatusWriter`). But `controllers/etcdcluster_controller.go` and its test are the only files both this PR and the updated base changed since the fork, so this will conflict on rebase and needs a careful manual reconcile.

Please rebase onto the current `fix/align_to_prev` tip (after its review fixes land), reconcile the controller file, and align the plugin TLS tests with the corrected detection. I'll re-review the e2e work against that base.

## Non-blocking

- **Scope/commit hygiene.** The `IsVoter` race fix and the `manager_auth_proxy` image bump (dead `gcr.io/kubebuilder` → `quay.io/brancz`) are both legitimate, but they're behavioral/infra changes riding inside a commit titled `feat(main): add e2e with kamaji case`. Split them into their own commits (or at least call them out in the PR description) so they bisect and review cleanly — the `IsVoter` fix in particular deserves to stand on its own.
- **Repeated literal.** `kamaji-e2e` is duplicated across `hack/e2e.sh`, the test, and the fixtures (documented as a sync hazard); consider sourcing it from one env var passed into the suite.

## What's genuinely good

The Kamaji test proves the full consumer story end to end: an operator-managed full-mTLS `EtcdCluster` backs a Kamaji `DataStore`, a `TenantControlPlane` comes up on it, a ConfigMap written through the *tenant* API is then found as a key in *our* etcd (via `etcdctl` exec). That validates the real contract rather than a proxy for it. `hack/e2e.sh` is careful too — diagnostics dumped from the EXIT trap before teardown (with a comment on why a post-step can't do it), pinned component versions with rationale, and the Kamaji webhook/datastore chicken-and-egg broken the same way the chart's own pre-install hook does. Build-tag-gated so it stays out of unit runs, and `e2e.yml` gates all PRs with a docs-only `paths-ignore`.

PR #324, opened 2026-06-05
feat(main): add e2e with kamaji case

androndo:


lllamnyp (approved):
Everything raised earlier is resolved.

- **Rebased onto `v1`** cleanly (merge-base is `v1`'s tip), so the stale fork point is gone.
- **Plugin TLS tests reconciled** to the corrected `--cert-file` detection — `helpers_test.go` now covers server-TLS-only, mTLS, no-cert-file, and mount-not-found via `findSecretNameForTLS`, with no overlap or helper-name collision against the `getTLSConfig` tests already on `v1`. Compiles and passes together.
- **Seed `IsVoter` fix** (`Status().Update` → `Status().Patch(MergeFrom)`) rebased cleanly and keeps its concurrent-writer regression test.
- **Workflow-trigger changes dropped.** I pushed a commit on top (`f8bbeec`, maintainer edit) reverting the `master`→`main` edits to `ci.yml` and `publish.yml` back to their `v1` state — they're pre-existing workflows unrelated to adding e2e, and `publish.yml` in particular would have armed a push to a Docker Hub repo the project no longer owns. The new `e2e.yml` is untouched and still gates PRs.

Build, `go vet`, the full unit suite, and `make manifests generate` (no drift) all pass on the rebased tree.

The e2e suite itself is the strong part: it proves the full consumer story end to end — an operator-managed full-mTLS `EtcdCluster` backs a Kamaji `DataStore`, a `TenantControlPlane` comes up on it, and a ConfigMap written through the tenant API is then found as a key in our etcd. The harness dumps diagnostics from the EXIT trap before teardown, pins component versions with rationale, and is build-tag-gated so it stays out of unit runs.

LGTM.

PR #325, opened 2026-06-05
fix(e2e): test waits rediness against flaky e2e

androndo:


lllamnyp (approved):
Fixes the e2e flake at its root. `etcdKeys` now selects a member pod by readiness — `Status.Phase == Running` with a `Ready` `etcd` container — instead of `pods.Items[0]`, and fails clearly if none exists. That's exactly the cause of the earlier `container not found ("etcd")`: the dump ran while a freshly-created member's container hadn't started. The comment captures the why (`Available` latches on quorum, not the full replica count; `etcdctl get` is linearizable, so any ready member returns the same keys).

Scope is test-only — one file, `kamaji_datastore_test.go`. Compiles under the `e2e` build tag, `vet` and full build pass, and the `kamaji-datastore` e2e check is green on this PR (~6m). Based cleanly on the current `v1`.

LGTM.

PR #326, opened 2026-06-05
fix(main): restore missed options by cozystack use-case

androndo:


lllamnyp (approved):
Ports the legacy free-form `spec.options map[string]string` to a closed, typed `EtcdOptions` (`quotaBackendBytes`, `autoCompactionMode`, `autoCompactionRetention`, `snapshotCount`), latches it through `status.observed`, propagates it to members at both creation sites, and renders it onto the etcd command line. The design choice to type the set rather than carry a free-form flag map — with new flags landing as new typed fields instead of an escape hatch — is the right call: it keeps the surface something the operator can reason about.

Verification:
- `go build ./...` and `go vet ./api/... ./controllers/...` clean.
- `make generate manifests` produces no drift — the checked-in deepcopy and CRD YAML match the types.
- Targeted tests pass: `TestBuildPod_AppliesEtcdOptions` (incl. the nil-Options "emits nothing" case), `TestSpecEqualsObserved_OptionsMatter`, `TestBootstrap_PropagatesOptionsToSeed`, `TestScaleUp_PropagatesOptionsToNewMember`, and the envtest `TestSchema_OptionsValidation`.

Correctness points checked:
- **Both member-creation sites carry Options** — `bootstrap` and `scaleUp` are the only non-test `EtcdMemberSpec{}` constructions, and both set `Options: cluster.Status.Observed.Options`. Replacement is delete-Pod-then-scaleUp, so there's no separate path that misses it.
- **Latching is complete** — `snapshotSpecIntoObserved` snapshots `Spec.Options`, and `specEqualsObserved` compares it; without the latter a mid-flight edit wouldn't re-snapshot (covered by `TestSpecEqualsObserved_OptionsMatter`).
- **No flag conflicts** — the four flags are emitted only by `optionFlags`, which guards every field so unset emits nothing; nothing else in `buildPod` sets them.
- **Schema validation is sound** — enum on the mode, `Minimum=0`/`Minimum=1` on the numerics (rejects `--snapshot-count=0`), and the retention pattern accepts a bare integer or a Go-duration shape.

Docs (`concepts.md` tuning-options section, `migration.md` map→typed section) match the implementation — field names, types, flag mappings, and the apply-on-create wording.

Non-blocking notes:
- **No cross-field check between `autoCompactionMode` and the retention format.** `mode: revision` with a duration retention (e.g. `"5m"`) passes schema validation but etcd rejects it at startup (revision mode expects an integer). etcd fails closed, so this is a footgun rather than a hazard; a CEL rule pairing the two could reject it pre-admission if you want it caught earlier. Fine to defer.
- The `autoCompactionRetention` comment says the pattern "admits what etcd itself parses"; etcd's parser also accepts `ns`/`us`/signed values, so the pattern is an intentional narrowing. Cosmetic — consider "the practical subset of what etcd parses."

LGTM.

PR #330, opened 2026-06-12
fix(main): add support peer-auto-tls

androndo:
## Summary by CodeRabbit

* **New Features**
  * EtcdMember supports a peer auto‑TLS mode and cluster adoption preserves legacy peer-auto-TLS.
  * Migration output now surfaces security warnings per-resource and provides a post‑apply security summary.

* **Documentation**
  * Added documentation describing peer auto‑TLS detection, migration behavior, and operational implications.

* **Bug Fixes**
  * Narrowed TLS peer validation error messaging for the exactly-one peer rule.

* **Tests**
  * Added tests covering peer auto‑TLS rendering, adoption, and precedence.

### DIFF
--- a/api/v1alpha2/cel_validation_test.go
+++ b/api/v1alpha2/cel_validation_test.go
@@ -477,7 +477,7 @@ func TestCEL_TLSPeerCertManagerAndSecretRefMutuallyExclusive(t *testing.T) {
 		_ = k8s.Delete(ctx, c)
 		t.Fatalf("apiserver accepted both peer.secretRef and peer.certManager; expected rejection")
 	}
-	if !strings.Contains(err.Error(), "exactly one of spec.tls.peer.secretRef or spec.tls.peer.certManager") {
+	if !strings.Contains(err.Error(), "exactly one of spec.tls.peer.secretRef") {
 		t.Fatalf("error did not mention peer mutual exclusion: %v", err)
 	}
 }
--- a/api/v1alpha2/etcdmember_types.go
+++ b/api/v1alpha2/etcdmember_types.go
@@ -44,6 +44,17 @@ type EtcdMemberTLS struct {
 	// mTLS (--peer-client-cert-auth=true).
 	// +optional
 	PeerSecretRef *corev1.LocalObjectReference `json:"peerSecretRef,omitempty"`
+
+	// PeerAutoTLS is operator-managed plumbing: it carries the cluster's
+	// reserved "etcd-operator.cozystack.io/peer-auto-tls" annotation down to
+	// the member so buildPod renders etcd's --peer-auto-tls (self-signed, no
+	// shared CA) instead of mounting a peer secret. INSECURE — peer is
+	// encrypted but NOT authenticated. Set only on clusters adopted from a
+	// legacy --peer-auto-tls cluster, and never together with PeerSecretRef
+	// (an explicit peer secret supersedes the annotation). Users do not set
+	// this directly; the cluster controller derives it.
+	// +optional
+	PeerAutoTLS bool `json:"peerAutoTLS,omitempty"`
 }
 
 // Condition types for EtcdMember.
--- a/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcdmembers.yaml
+++ b/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcdmembers.yaml
@@ -1336,6 +1336,17 @@ spec:
                         type: string
                     type: object
                     x-kubernetes-map-type: atomic
+                  peerAutoTLS:
+                    description: |-
+                      PeerAutoTLS is operator-managed plumbing: it carries the cluster's
+                      reserved "etcd-operator.cozystack.io/peer-auto-tls" annotation down to
+                      the member so buildPod renders etcd's --peer-auto-tls (self-signed, no
+                      shared CA) instead of mounting a peer secret. INSECURE — peer is
+                      encrypted but NOT authenticated. Set only on clusters adopted from a
+                      legacy --peer-auto-tls cluster, and never together with PeerSecretRef
+                      (an explicit peer secret supersedes the annotation). Users do not set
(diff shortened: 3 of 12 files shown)


lllamnyp (changes requested):
The capability is the right one — adopting a legacy `--peer-auto-tls` cluster needs replacement/scaled members to speak the same insecure peer mode, or they can't rejoin. Three changes before merge; the first is an API-shape consistency point, the other two are about making an insecure mode impossible to adopt unknowingly.

## 1. Don't expose this on the typed `spec` — make it a reserved annotation

`spec.tls.peer.autoTLS` adds a **user-settable, first-class, insecure** peer-TLS mode sitting right next to `secretRef` and `certManager`. Its only sanctioned use is legacy adoption (the field doc says so), yet nothing stops someone picking it for a brand-new cluster and getting an unauthenticated peer plane. "Migration-only" is enforced by a comment, not the API.

This is the same call we made for `headlessServiceName` / `dataDirSubPath`: migration-only knobs a user should never choose belong on a **reserved annotation the operator interprets**, not the typed spec. The mechanics differ here — peer-auto-tls is cluster-wide and must propagate to every member the operator creates (no shared CA ⇒ a replacement *must* also run `--peer-auto-tls`), so it doesn't self-wipe like those two. So the right home is a reserved **cluster-level** annotation, e.g. `etcd-operator.cozystack.io/peer-auto-tls`, that:

- the cluster controller reads (in `deriveMemberTLS`, in place of `spec.tls.peer.autoTLS`) and propagates to every member it builds;
- `buildPod` consumes (the existing internal member-level signal is fine as operator-managed plumbing) to render `--peer-auto-tls`;
- takes effect **only when no explicit peer `secretRef`/`certManager` is set**, and is ignored/superseded otherwise (precedence has to live in code now, see below).

Then drop `PeerAutoTLS` from the public types and **revert the CEL rule** back to the two-way `has(secretRef) != has(certManager)` — auto-tls is no longer part of the typed union.

Acknowledged tradeoff (and it's the point): as an annotation you can't CEL-validate exclusivity or hard-prevent a user from setting it. That's acceptable — an undocumented reserved key is far less of a footgun than a discoverable typed `autoTLS` option, and it keeps the public TLS API to the two real, secure modes.

## 2. Document it in `docs/migration.md`

This PR ships an entire insecure peer mode plus migration carry-forward with **no doc coverage** (`migration.md` doesn't mention peer-auto-tls at all). Add a subsection under the TLS material covering:

- what `--peer-auto-tls` is — per-member self-signed certs, **no shared CA, encrypted but not authenticated** (any TLS-capable workload reaching `:2380` can peer);
- that adoption **detects and carries it forward** (now via the reserved annotation) so replacement/scale keeps working — and *why* (no CA exists to do real mTLS, a plaintext-peer replacement could never join);
- that the only off-ramp to real mTLS is **delete-and-recreate** (or the careful manual rolling restart, given persistent storage — strict-mTLS and auto-tls members can't peer, so there's a brief no-quorum window at the cutover);
- that this is a *distinct* scenario from the explicit-mTLS SAN coverage issue already documented — don't conflate them.

## 3. Make the warning loud

The `adopt.go` warning is correct in content but renders identically to routine `  warning:` lines in the plan and is easy to miss. For a **security-posture downgrade**, elevate it: a distinct marker (e.g. a `SECURITY:` / `⚠` callout), and re-surface it in the **post-`--apply` summary**, not only in the pre-apply plan — so an operator cannot finish a migration without seeing that they just adopted an unauthenticated peer plane. (Update its `spec.tls.peer.autoTLS` references to the new annotation while you're there.)

lllamnyp (approved):
All three change-request items are addressed, cleanly. Approving.

- **Off the typed spec → reserved annotation.** `spec.tls.peer.autoTLS` / the `PeerAutoTLS` struct are gone from the public types (the `PeerTLS` CEL rule is back to the two-way `secretRef != certManager`), replaced by the reserved cluster annotation `etcd-operator.cozystack.io/peer-auto-tls`. Precedence is enforced in code: `clusterPeerAutoTLS` returns false whenever any explicit `spec.tls.peer` is set, so real mTLS always wins and the annotation only applies when there's no typed peer TLS.
- **Documented.** New `docs/migration.md` "Peer auto-TLS" section — what it is (per-member self-signed, no shared CA, encrypted-not-authenticated), that adoption carries it forward via the annotation to keep replacement/scale working and why, and that the off-ramp to real mTLS is delete-and-recreate.
- **Loud warning.** A distinct `SecurityWarnings` channel (separate from routine `Warnings`), rendered inline as `⚠️ SECURITY:` and re-surfaced after `--apply` via `renderSecuritySummary` — so an unauthenticated-peer adoption can't be missed.

What I checked beyond that:

- **Peer-URL scheme consistency** (the main correctness risk): both `memberPeerScheme` and `clusterPeerScheme` return `https` for the auto-tls case, and the seed / pending / existing peer URLs all route through `clusterPeerScheme`, so advertised URLs, `--initial-cluster` entries, and the listener in `buildPod` agree. A mismatch here would prevent quorum; it's consistent.
- **No nil-deref in `buildPod`** — the new `switch` takes the `PeerAutoTLS` branch before `peerTLS`, and the `PeerSecretRef` deref only runs when it's non-nil.
- **Mutual exclusion** of `PeerAutoTLS` vs `PeerSecretRef` holds by construction in both `deriveMemberTLS` and `deriveAdoptedMemberTLS`.
- **Detection** stamps the annotation only when no peer TLS is declared and a live member advertises `https://` — matching the legacy operator's unconditional-`--peer-auto-tls`-unless-`PeerSecret` behavior.
- Build, full test suite, and `make manifests generate` (no drift) all pass.

One non-blocking, cosmetic nit: `api/v1alpha2/cel_validation_test.go:480` loosened its assertion from the full CEL message to a prefix. The rule message in `etcdcluster_types.go` is unchanged and still carries the full string, so the original full-message match would still pass — this loosening is unrelated to the feature and weakens the test slightly for no gain. Worth reverting to the full-message assertion, but not a blocker.

LGTM.

PR #336, opened 2026-06-15
fix(controllers): replace a persistently crash-looping member

androndo:
A non-bootstrap PVC member whose etcd cannot start — classically because its data dir was lost while the cluster membership moved on, leaving its frozen --initial-cluster stale (etcd refuses to boot: "error validating peerURLs ... member count is unequal") — crash-loops forever with no recovery path. The existing self-heal covers only memory-medium members (pod lost => data lost); for PVC members the PVC survives pod restarts, so "pod lost" != "data lost" and the stale --initial-cluster is never escaped.

Detect such a member (etcd container not ready and restarted past a threshold, excluding OOMKilled) and delete it for replacement — but only when the rest of the cluster still has quorum, so a cluster-wide outage never cascades into mass deletion (the finalizer's MemberRemove is quorum-gated too). The cluster controller then gap-fills a fresh member with a current --initial-cluster.

Also extend the Kamaji e2e to wait for readyMembers==3, then churn all three members one at a time and re-verify the tenant API still roundtrips through the fully replaced (hash-named) member set — guarding that member naming/replacement stays transparent to a Kamaji DataStore (stable <cluster>-client Service + wildcard server-cert SAN).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added self-healing for persistently crash-looping PVC-backed etcd members, with quorum gating to prevent cascading replacements.
  * Excludes bootstrap members from automatic replacement and avoids triggering recovery for OOMKilled cases.
* **Tests**
  * Expanded unit and end-to-end coverage for crash-loop self-heal behavior and member churn.
  * Increased e2e test timeout to improve stability.
* **Documentation**
  * Updated operational runbooks and concepts to clarify PVC crash-loop replacement behavior, timing expectations, and quorum safeguards.


lllamnyp (changes requested):
## What this change does

Adds self-heal for PVC-backed members: in `updateStatus`, when a non-bootstrap, non-memory member's etcd container is "stuck" (not ready, `RestartCount >= 5`, last termination not `OOMKilled`) AND the rest of the cluster has quorum, the operator deletes the `EtcdMember` CR. The finalizer then runs `MemberRemove`, the member-owned PVC is GC'd, and the cluster controller gap-fills a fresh `GenerateName` member with an empty data dir. The mechanism mirrors the existing memory-backed pod-loss self-heal and is sound: PVC owner-ref GC discards the corrupt data dir, and the quorum gate plus write-requires-quorum on `MemberRemove` prevent a cascade. Build, `go vet`, and the full `controllers` unit suite pass; the e2e file compiles under `-tags e2e`.

The code logic is correct. The blockers below are about documentation that now contradicts the code, and about the absence of end-to-end coverage for the new path.

## Blocking issues

### 1. Documentation directly contradicts the new behavior

The repo documents — repeatedly and emphatically — that PVC-backed members do NOT auto-replace. This PR introduces exactly that auto-replacement and updates none of it. Since the PR changes this behavior area, it owns these docs.

- **`README.md:35`** — states verbatim: *"No automatic broken-member replacement for PVC-backed clusters (memory-backed members do auto-replace on Pod loss; `status.brokenMembers` reads 0 in practice ...)"*. Now false. Rewrite to describe the new crash-loop self-heal (non-bootstrap PVC member, `RestartCount >= dataLossRestartThreshold`, OOMKilled excluded, quorum-gated) and drop/qualify the "No automatic broken-member replacement" claim.

- **`docs/concepts.md:130`** — the storage table's PVC "Pod loss →" cell reads *"Same Pod / new Pod re-uses existing data dir; etcd rejoins with the same member ID and `ClusterID`."* It omits the new terminal case: a PVC member whose data dir is lost/corrupt and that crash-loops past the threshold is now deleted and replaced with a fresh member ID (not a same-ID rejoin). Add the crash-loop replacement case to the PVC row.

- **`docs/concepts.md:156`** — states *"For PVC-backed members today, `isBroken` stays a stub; richer detection is a future concern."* and frames PVC corruption replacement as *"a future hook ... (e.g. PVC corruption with a grace period)."* That "future concern" is now implemented (via `etcdContainerStuck`, not via `isBroken`/`BrokenMembers`). Update to reflect that PVC crash-loop self-heal now exists, and clarify it does not flow through `BrokenMembers`.

- **`docs/operations.md:462-469`** ("Pod loss and auto-replacement") describes auto-replacement as a memory-only / Pod-loss mechanism keyed on `Status.PodUID`. There is now a second, PVC-specific replacement trigger (persistent crash-loop) with different detection (`RestartCount`, container readiness) and different latency (≥5 restarts at the up-to-5m CrashLoopBackOff cap ⇒ tens of minutes, not ~5s). Add a subsection documenting the PVC crash-loop replacement path — threshold, OOMKilled exclusion, quorum gate, and expected detection latency — so operators don't mistake a deliberately-deleted member for operator misbehavior.

### 2. The new e2e test does not exercise the new code path

`test/e2e/kamaji_datastore_test.go` adds a "member churn" block that deletes each member via `kube.Delete` (the normal scale-down / `MemberRemove` path) and asserts Kamaji keeps working through `GenerateName`-hashed replacements. This is a good churn test, but it never triggers `etcdContainerStuck` / the crash-loop self-heal — no member is made to crash-loop, no `RestartCount` is driven past the threshold. The production logic added by this PR therefore has zero e2e coverage; only the fake-client unit tests (`TestEtcdContainerStuck`, `TestUpdateStatus_ReplacesStuckMember`, `TestUpdateStatus_KeepsStuckMemberWithoutQuorum`, `TestUpdateStatus_KeepsStuckBootstrapMember`) cover it.

This is also the test that satisfies issue 1's contract: the docs claim PVC members auto-replace on persistent crash-loop, and that contract is between the running operator and a running cluster, so it belongs in e2e, not in a doc-grepping test. Add an e2e case that induces a real persistent crash-loop on a PVC-backed member (e.g. corrupt/empty the data dir so etcd hits "member count is unequal", or otherwise force `RestartCount` past `dataLossRestartThreshold` while the other two members stay ready) and assert: (a) the stuck member's CR is deleted, (b) a fresh `GenerateName` member with a new member ID appears, (c) the old `data-<name>` PVC is GC'd, and (d) the cluster returns to 3 ready members and still serves reads/writes.

## Analyzed, not a blocker

**Quorum threshold (`controllers/etcdmember_controller.go:954`).** `clusterHasQuorumWithout` returns `ReadyMembers >= desired/2+1`. In Go integer arithmetic `desired/2+1 == floor(desired/2)+1`, which is exactly etcd quorum for every N (odd and even). The gate counts ready OTHERS (the stuck member is `!podReady`, so excluded from `ReadyMembers`), which is the correct set for the `MemberRemove` that runs while the stuck member is still registered. No change required. Optional: extend the unit table to cover desired ∈ {1,2,4,5}; today only desired=3 at ready ∈ {1,2} is covered.

## Non-blocking observations

- `clusterHasQuorumWithout` (`controllers/etcdmember_controller.go:937-954`) reads `cluster.Status.ReadyMembers`, set by the cluster controller, which can lag. Worst case in a multi-member outage, a stale-high count lets one member be deleted before the count catches up; the next member's reconcile sees the decremented count and is kept, and the finalizer's `MemberRemove` fails closed without quorum. Acceptable degradation (matches the documented memory behavior), but worth a line in issue 1's new docs subsection.

- `dataLossRestartThreshold = 5` is reasonable, but a slow restore or slow learner join on a replacement member could trip it, producing repeated replacement. Quorum-gated so it can't break the cluster, and a genuinely-stuck replacement should be replaced — acceptable, but note it in the new docs so operators expect it.

lllamnyp (approved):
All previously raised concerns are addressed. Approving.

## Verification performed

- `go build ./...`, `go vet ./controllers/`, `go vet -tags e2e ./test/e2e/` — all clean.
- `go test ./controllers/ -count=1` — pass, including the four new self-heal unit tests and `TestEtcdContainerStuck`.

## How the earlier blockers were resolved

**Documentation.** The stale "no automatic broken-member replacement for PVC-backed clusters" claim is gone. `README.md:35`, the `docs/concepts.md` storage table, and the new **Crash-loop self-heal (PVC members)** section now describe the actual behavior — trigger (`RestartCount >= 5`), OOMKilled exclusion, quorum gate, tens-of-minutes latency, and self-limiting replacement-of-replacement. The new section anchor matches the GitHub slug and the storage-table cross-link resolves. `docs/operations.md` gains a matching subsection. The `status.brokenMembers` "reads 0" claim is preserved and correctly explained (self-heal triggers off restart count, not `isBroken`).

**End-to-end coverage.** `test/e2e/member_selfheal_test.go` (`TestPVCMemberCrashLoopSelfHeal`) induces a real crash-loop by corrupting the victim's data dir, then asserts the full chain: stuck member CR deleted, `data-<member>` PVC GC'd, a fresh `GenerateName` member gap-fills, cluster back to 3 ready, and etcdctl put/get still serves. Makefile e2e timeout raised 30m→45m for the CrashLoopBackOff ramp.

## Correctness trace (sound)

- **Quorum math** (`controllers/etcdmember_controller.go:979`): `readyOthers >= desired/2+1`; integer `desired/2+1 == floor(desired/2)+1 ==` etcd quorum for all N. Conservative.
- **Stale-count compensation** (`:972-978`): if this member's own stored `MemberReady` condition is still `True`, it is subtracted from `ReadyMembers`, so a lagging stale-high count cannot green-light a delete that drops below quorum. Covered by `TestUpdateStatus_KeepsStuckMemberWhenStaleReadyCountIncludesIt`.
- **No stale write after delete** (`:1044-1054`): the self-heal branch returns immediately after `r.Delete`, never reaching `r.Status().Update`.
- **Replacement serialized**: the cluster controller blocks further scaling while a member deletion is in flight, then gap-fills via `scaleUp`. The finalizer's `MemberRemove` is independently quorum-gated (belt-and-braces).
- **`etcdContainerStuck`** (`:924-944`): excludes `DeletionTimestamp != nil`, `Ready`, `RestartCount < 5`, and `OOMKilled` in both current and last termination state; bootstrap/memory members excluded at the call site. Exhaustively covered by `TestEtcdContainerStuck`.
- **PVC GC**: the PVC carries a controller owner-ref to the member, so deletion discards the corrupt data dir.

## Residual note (not blocking)

The stale-count compensation handles "condition still `True` but no longer ready." The opposite narrow transient — condition already `False` while `ReadyMembers` is still stale-high by one — is not compensated by the subtraction, so the gate could briefly read one too high. This is already defended by the independent quorum gate on the finalizer's `MemberRemove`, and the e2e test exercises real (non-fake-client) timing. Flagged for awareness only; no change required.

PR #337, opened 2026-06-18
feat(main): add image parameters for air-gapped cases

androndo:
## Release Notes
* **New Features**
  * Added `EtcdCluster.spec.imagePullSecrets` to inject image pull secrets into member Pods (including the restore initContainer) and latched them via `status.observed`.
  * Added `EtcdImage` repository defaulting: configure an operator-wide etcd image repository via Helm `etcdImage.repository` / `--etcd-image-repository` when members don’t specify an image repository.
* **Bug Fixes**
  * Reconciliation and in-place adoption now propagate pull-secret changes to newly created and updated members.
* **Documentation**
  * Updated installation/migration notes to describe the new mirror and pull-secret behavior.
* **Tests**
  * Added unit and e2e coverage for image resolution and pull-secret propagation.


lllamnyp (changes requested):
**Request changes — narrow this to the operator-wide repository flag (+ pull secrets)**

Thanks for tackling the air-gap case. The operator-wide repository default is the right primitive and I want to merge that part. But the per-cluster `spec.image` block introduces a second source of truth for the etcd version that can silently disagree with `spec.version`, and that's not something I'm willing to take on. Concretely:

**Keep:**
- `--etcd-image-repository` / `ETCD_IMAGE_REPOSITORY` on the binary, the chart's `etcdImage.repository` value, and the env wiring in the Deployment.
- `spec.imagePullSecrets` (cluster → member mirror, `status.observed` latching, and the migrate carry-through). A private mirror needs credentials, and pull secrets have none of the version-ambiguity problem below — they only affect *whether* the pull authenticates, never *what version* runs.

The repository flag cleanly serves the mirror-once-per-fleet case: it only changes *where* the image is pulled from, never *what version* runs.

**Please drop (for now):**
- `EtcdCluster.spec.image` (`EtcdImageSpec` — `repository`, `tag`, `pullPolicy`) and its mirror on `EtcdMember.spec.image`
- the `status.observed.image` latching and the `resolveEtcdImage` tag/pull-policy handling
- the migrate `etcdImageOverride` reconstruction

The blocking problem is `spec.image.tag`. The operator treats `spec.version` as the source of truth for the running etcd version — it's injected as `ETCD_VERSION` and drives the restore version-compat pre-flight, the latched target, and drift detection. `spec.image.tag` feeds none of that; it only changes the pulled reference. So a spec where `image.tag` resolves to a different minor than `spec.version` is internally contradictory: the container runs one version while the operator reasons about another. Nothing validates that they agree, and the most damaging consequence lands exactly on the air-gap path this PR targets — the restore initContainer rebuilds the data dir to `spec.version`'s format, then a mismatched binary boots on it. Adding a second, unvalidated way to set the version is a footgun I'd rather not introduce, even guarded by docs.

---

**Future work (not blocking — please open issues, don't fold into this PR):**

1. **Observed member version in `status`.** `spec.version` communicates intent but is currently relied on as the source of truth. The member should determine the etcd version it's actually running at runtime and write it to its `status`; the operator's version-dependent logic should key off that observed value, not the spec field. This is the principled fix for the conflict above and would make a version override safe to reconsider later.

2. **Multi-version restore agent.** The restore agent ships a single compiled `etcdutl` (3.6.x in this build), so restore is silently pinned to the operator's own etcd minor regardless of `spec.version`. If we want the operator to support running multiple etcd versions, the restore agent needs to run a matching `etcdutl` per target version too — otherwise restore is unsupported for any cluster off the operator's minor.

lllamnyp (approved):
**Approving** — the strip resolves the blocking concern. Thanks for the quick turnaround.

What I checked:
- `spec.image` (`EtcdImageSpec` — repository/tag/pullPolicy) is fully removed, including the `EtcdMember` mirror, the `status.observed.image` latching, deepcopy, and the CRD YAML. No dangling references remain.
- `resolveEtcdImage` now always pins the image to `<repo>:v<spec.version>`. There is no longer any way to set a tag that disagrees with `spec.version`, so the version-conflict footgun is gone by construction rather than by documentation.
- `spec.imagePullSecrets` is retained (cluster → member mirror, `status.observed` latching, migrate carry-through) — a private mirror still gets its credentials, and pull secrets carry no version ambiguity.
- Operator-wide `--etcd-image-repository` / `ETCD_IMAGE_REPOSITORY` and the chart's `etcdImage.repository` are retained.

Build is clean and the migrate + controller tests pass.

**Non-blocking follow-up (please address in a separate small change, your call on timing):**

The migrate path now consumes only the *tag* of a legacy etcd image (via `extractVersion`) and silently discards the registry. A cluster that ran a private mirror — e.g. `registry.internal/mirror/etcd:v3.6.11` — translates to an `EtcdCluster` that resolves against the operator default registry, with no entry in the dropped list and no warning. The members then `ImagePullBackOff` until the operator is separately configured with `--etcd-image-repository`, which is exactly the air-gap audience this PR targets.

The fix is not to reintroduce a per-cluster image override — it's a migration **warning** emitted when the legacy etcd image's registry differs from the operator default, pointing the user at `--etcd-image-repository`. The migration is the one place where "this cluster used a non-default registry" is known, and right now that information is dropped silently.

**Future work captured in my earlier review (issues, not this PR):** observed member version written to `status` rather than relying on `spec.version` as source of truth; and a multi-version-capable restore agent so restore isn't pinned to the operator's single compiled `etcdutl` minor.

Issue #338, opened 2026-06-29
Observe the running etcd version in EtcdMember status and key version-dependent logic off it


androndo:
## Background

Spun out of the review on #337 (per reviewer request — future work, not folded into that PR).

Today `spec.version` is treated as the **source of truth** for the running etcd version: it is injected as `ETCD_VERSION`, drives the restore version-compat pre-flight, the latched target, and drift detection. The member never reports what version it is *actually* running.

## Problem

Because nothing observes the real running version, intent (`spec.version`) and reality can silently diverge — there is no signal to detect or reconcile a mismatch. This was the blocking reason a per-cluster image tag override (`spec.image.tag`) was dropped from #337: a tag resolving to a different minor than `spec.version` is internally contradictory and unvalidated.

## Proposal

- The member determines, at runtime, the etcd version it is actually running and writes it to its `status` (e.g. `EtcdMember.status.version`, read from the etcd endpoint / `/version`).
- The operator's version-dependent logic keys off that **observed** value, not the spec field.

## Why it matters

This is the principled fix for the spec-vs-reality conflict. With an observed version in status, a per-cluster image/version override (deferred in #337) could be reconsidered safely, and version drift/upgrade handling becomes detectable rather than assumed.

Refs: #337

PR #340, opened 2026-07-03
feat(main): add observed versions of EtcdMembers

androndo:
Observe the etcd version each member is actually running (via the Maintenance Status RPC against the member's own endpoint) into EtcdMember.status.version, surfaced as the "Running" print column. Add an informational VersionDrifted condition comparing the observed running version against the intended spec.version; it is left unset when intent is not yet known. Observation is best-effort and never gates readiness. Extract the operator-side dial config out of discoverMemberID into a shared etcdDialConfig helper.

Implements #338.

## Summary by CodeRabbit

* **New Features**
  * Added visibility into the actual etcd version running on each member.
  * Exposed a new status field and kubectl column so cluster state is easier to inspect at a glance.
  * Added an informational version-drift condition to show whether the running version matches the desired version.

* **Bug Fixes**
  * Improved status reporting so version checks are best-effort and won’t block member readiness.
  * Preserved the last known version when live version checks fail.

### DIFF
--- a/api/v1alpha2/etcdmember_types.go
+++ b/api/v1alpha2/etcdmember_types.go
@@ -63,6 +63,12 @@ const (
 	MemberJoined = "Joined"
 	// MemberReady indicates the member is healthy and serving requests.
 	MemberReady = "Ready"
+	// MemberVersionDrifted is True when the version etcd actually reports
+	// running (status.version, observed from the endpoint) does not match the
+	// version the operator asked this member to run (spec.version). It makes
+	// intent-vs-reality version drift detectable rather than assumed; the
+	// operator does not act on it (spec.version still drives the image tag).
+	MemberVersionDrifted = "VersionDrifted"
 )
 
 // EtcdMemberSpec defines the desired state of a single etcd member.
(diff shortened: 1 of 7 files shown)


lllamnyp (approved):
LGTM — approving.

Adds an observed running-version field (`EtcdMember.status.version`), a `VersionDrifted` condition, a `Running` print column, and refactors the operator-side dial config out of `discoverMemberID` into a shared `etcdDialConfig`. The change is careful and well-tested: `go build ./...`, `go vet ./...`, and `go test ./controllers/...` all pass, `make generate manifests` produces no drift (CRD, deepcopy, print columns in sync), and the docs/concepts.md claims check out against the code. All five new tests exercise both ready paths (discovery branch with `MemberID==""` and the steady-state `default` branch): version match → drift False/VersionMatched, mismatch → drift True/VersionMismatch, empty intent → condition unset, drift resolution, and Status-RPC error being non-fatal. Endpoint selection is correct — `observeVersion` dials only `self` (per-endpoint Status), unlike `discoverMemberID` which routes through voter peers. Deletion is guarded (Reconcile returns early on DeletionTimestamp before `updateStatus`). Version formats align: `spec.version` is bare and etcd `StatusResponse.Version` is bare too, so the direct string comparison is correct.

Two non-blocking recommendations:

1. **Recurring steady-state cost** (`controllers/etcdmember_controller.go:1104-1108`). Before this change, a Ready member's 30 s reconcile hitting the `default` branch did no etcd I/O. Now every Ready reconcile opens a fresh etcd gRPC client (TLS handshake + auth) and issues a Status RPC, and for TLS clusters `etcdDialConfig` additionally fetches the parent EtcdCluster plus the CA/client/credential secrets each time. That is N dials + several secret reads per 30 s cluster-wide, forever, purely for an informational field. It is bounded (5 s timeout) and best-effort, so not a defect, but consider observing less often than every reconcile if fleet load matters.

2. **Empty-version clobber guard** (`controllers/etcdmember_controller.go:1105`). `if v := r.observeVersion(...); v != member.Status.Version` will overwrite a good `status.version` with `""` if a *successful* Status ever returns an empty `Version`. Real etcd always populates it, so this is not currently reachable, but tightening to `if v != "" && v != member.Status.Version` is cheap hardening (cover with a test: fake with empty `statusVersion` and nil error, assert the prior value survives).

PR #347, opened 2026-07-28
feat(controllers): stop cascade-deleting data volumes with their member

kvaps:
## Problem

A member's data PVC carries a controller owner reference to its `EtcdMember`. That reference does two jobs at once: it marks the owner, and it makes the volume cascade-delete with it. The first is wanted. The second means every route that removes an `EtcdMember` destroys the data behind it — irreversibly, on any `Delete`-reclaim StorageClass:

- `kubectl delete etcdmember` by someone debugging;
- a GitOps prune;
- a CRD replacement during an upgrade;
- a chart rollback, or any `helm upgrade` where the CRs stop being rendered.

None of those events say anything about whether the data still matters, yet all of them are fatal to it.

This is not hypothetical. On a dev cluster the `EtcdMember` CRs of seven namespaces were removed after a migration; the volumes went with them, and the running pods kept serving from already-deleted volumes until a routine node reboot ended them. The data had been unrecoverable for twelve days by then, with nothing in the API indicating it.

## Approach

Separate identity from lifetime, following the model CloudNativePG uses for its instance volumes ([`pkg/reconciler/persistentvolumeclaim`](https://github.com/cloudnative-pg/cloudnative-pg/tree/main/pkg/reconciler/persistentvolumeclaim)): PVCs carry no owner reference, and deletion happens where the operator explicitly decides it — `EnsureInstancePVCGroupIsDeleted`, called from scale-down and from the unschedulable-pod path.

**Identity** → `etcd-operator.cozystack.io/member-uid`. The member controller mounts a volume only when the annotation matches its own UID; anything else is refused rather than adopted (inheriting a foreign data dir crash-loops the pod once etcd finds a removed member ID in the WAL). Volumes created by older builds, and those stamped by `cmd/etcd-migrate`, carry a controller owner reference and no annotation — the first reconcile after upgrading migrates them in place: annotation stamped, owner reference stripped. Existing refusal semantics are unchanged, and the existing `TestEnsurePVC_Refuses*` tests still pass untouched.

**Lifetime** → three explicit paths, each a place the operator itself decided the data is finished:

| path | why deleting is right |
|---|---|
| scale-down (`scaleDown`) | the departing member's contents live on the remaining peers — its finalizer ran `MemberRemove` first. The shrink was requested, so storage returns to the pool. |
| crash-loop replacement (`updateStatus`) | the member cannot boot on its own data dir, and the existing quorum gate proved the rest of the cluster is healthy. The replacement starts clean and syncs from peers. Leaving the volume would also strand it — replacements get fresh names, so nothing would mount it again. |
| cluster deletion (`handleClusterDeletion`) | deleting the `EtcdCluster` is the user declaring the cluster finished, so its storage goes with it. |

Cluster deletion runs from a new finalizer, `etcd-operator.cozystack.io/cluster-cleanup`, held until the volumes are **actually gone**, not merely asked to go: a volume stays `Terminating` while a Pod still mounts it, so releasing early would let the `EtcdCluster` disappear with storage still allocated behind it — and would let a namespace delete look complete while PVCs were still draining. Reclaiming is scoped by the cluster label, so a second cluster sharing the namespace is untouched.

Everything else leaves the volume in place: a stray `kubectl delete etcdmember`, a GitOps prune, a CRD replacement, a chart rollback. Those say nothing about whether the data still matters; deleting the cluster does.

**Visibility** → `etcd-operator.cozystack.io/pvc-status`, mirroring `cnpg.io/pvcStatus`: `initializing` / `ready` / `detached`. Nothing branches on it; it turns "is this volume in use or left behind?" into a `kubectl get pvc -o custom-columns` instead of a cross-reference against the member list. A paused cluster's volume reads `detached`, which is also what a leftover reads — both are intact and unmounted, which is exactly the truth.

## Tests

New, all failing against the current behaviour where applicable:

- `TestEnsurePVC_CreatesVolumeWithoutOwnerReference` — no owner reference, marker + `initializing` stamped, and the marker alone is enough to recognise the volume. *(fails on main: volume comes back owner-referenced)*
- `TestEnsurePVC_MigratesLegacyOwnerReference` — a legacy volume gets its marker and loses the reference.
- `TestMemberDeletion_LeavesTheVolumeBehind` — the finalizer path removes the member from etcd and leaves the volume untouched.
- `TestScaleDown_ReclaimsTheVictimsVolume` — the victim's volume is reclaimed, the survivors' are not. *(fails on main: nothing reclaims it explicitly)*
- `TestScaleDown_PauseKeepsTheVolume` — 1→0 parks the member dormant; the volume stays.
- `TestUpdateStatus_ReplacingStuckMemberReclaimsItsVolume` — replacement reclaims the broken volume. *(fails on main)*
- `TestUpdateStatus_KeepingStuckMemberKeepsItsVolume` — when the quorum gate refuses the replacement, the volume is untouched: a cluster-wide outage looks identical on every member, and reclaiming there would turn an outage into data loss.
- `TestClusterDeletion_ReclaimsVolumesBeforeReleasingFinalizer` — volumes are removed, and the finalizer is only released on the pass where none remain.
- `TestClusterDeletion_HoldsFinalizerWhileAVolumeDrains` — a `Terminating` volume keeps the finalizer held.
- `TestClusterDeletion_LeavesAnotherClustersVolumes` — the reclaim is label-scoped; a co-tenant cluster's data is untouched.

The pause/resume tests were updated to assert the marker instead of the owner reference, and now also check that a paused volume reads `detached`.

`go test ./...` and `go vet ./...` are green.

## Docs

New `Data volume lifecycle` section in `docs/concepts.md` covering the rationale, the three deletion paths, and the status annotation. Updated every place that promised owner-ref GC of volumes: the API model, scale-to-zero, crash-loop self-heal, scale-down and pause in `docs/operations.md`, the cluster-deletion note in `docs/installation.md`, and the recreate-after-delete recipe (the cluster finalizer clears the volumes, so wait for the `EtcdCluster` to actually disappear before re-applying).

RBAC: the cluster controller gains `update` on `etcdclusters` (for the finalizer) and `list`/`delete` on `persistentvolumeclaims`; `charts/etcd-operator/files/manager-role-rules.yaml` regenerated.

## Summary by CodeRabbit

- **New Features**
  - Enhanced data volume lifecycle handling during scale-down, pause-to-zero, and member replacement, including explicit reclamation and safer PVC retention.
  - PVC lifecycle status is now exposed via annotations: initializing, ready, and detached.
  - Legacy PVC ownership is migrated automatically for smooth transitions.

- **Bug Fixes**
  - Cluster deletion now properly holds cleanup until related PVCs finish draining, then releases the finalizer.

- **Documentation**
  - Updated lifecycle and recovery guidance for teardown, pause behavior, and crash-loop replacement.

- **Tests**
  - Expanded coverage for PVC retention/reclaim semantics and finalizer behavior.


lllamnyp (changes requested):
**NOT LGTM** — on direction first, and separately on correctness. To be clear up front: the craftsmanship is not the problem. The tests are good, the docs are thorough, the CNPG precedent is real, and the pause/dormant semantics are correctly preserved. The problem is that this PR spends a large, irreversible behavioral change to buy *partial* protection against an event that can be prevented outright, at the API boundary, with a few dozen lines of chart YAML and a one-line docs fix.

## Direction: this moves the danger, it doesn't remove it

**1. Survival is not recovery.** After the flagship scenario — someone deletes the `EtcdMember` CRs — the volumes now survive, but by this PR's own design nothing can ever mount them again: replacements get fresh apiserver-assigned names and UIDs, and `pvcBelongsTo` refuses a foreign marker (correctly — inheriting a stale data dir crash-loops the pod). Meanwhile the member finalizers still run `MemberRemove` one by one, so the etcd cluster still dismembers itself, and (with #343) parks in `AllMembersLost`. What's left is archaeology material: intact bytes that require manual surgery (mount volume in a pod, `etcdutl` snapshot, restore into a new cluster) to become a database again. That's much better than destroyed bytes — but the incident still happens, the cluster still dies, and the recovery is still human-hours. The event itself remains one unguarded `kubectl delete` away.

**2. The threat model is applied inconsistently, one level up.** The PR argues that "a chart rollback, or any `helm upgrade` where the CRs stop being rendered … say[s] nothing about whether the data still matters" — when the object removed is an `EtcdMember`. But the exact same Helm event removing the **`EtcdCluster`** is treated as fully informed consent to destroy everything, and the new `cluster-cleanup` finalizer makes that path *more* thorough than today: the label-scoped sweep in `handleClusterDeletion` also catches detached leftover volumes that current `main` would leave behind (once the members are gone there is no owner chain to cascade through). Concretely: after a member-deletion accident, the reflex move — and the documented recovery for the no-data case in #343's runbook — is "delete the `EtcdCluster` and recreate it". Under this PR, that move erases the surviving volumes this PR just saved. We'd be trading protection against member-level accidents for amplification of cluster-level ones, on the one object users *are* expected to touch.

**3. The costs are structural, not incidental.**
- `adoptLegacyPVC` strips owner references fleet-wide on the first reconcile after upgrade, automatically and irreversibly. A subsequently *downgraded* operator refuses every annotation-only volume (`pvcOwnedBy` requires the owner ref), so after any node event an old operator cannot bring pods back up. That's a one-way door with no gate, no opt-out, and no mention in docs or release notes.
- The operator gains `list`/`delete` on PVCs plus a new body of deletion code — and the most destructive loop in it currently selects victims by a user-attachable grouping label (see blockers below). The failure mode moves from "human deletes the wrong CR" to "operator bug deletes the volumes", which is not obviously an improvement.
- The `EtcdCluster` finalizer adds the classic stuck-namespace mode when the operator is uninstalled first (the existing docs recovery for exactly that situation is now incomplete — see blockers).
- CNPG's model is imported without CNPG's counterpart: CNPG can re-attach instance PVCs to new instances; here a leftover volume is unmountable forever by design.

**What I'd do instead, first:** enforce the contract we've documented since day one — `docs/concepts.md` has said from its first version that `EtcdMember` is created and deleted by the cluster controller and users shouldn't touch it. A `ValidatingAdmissionPolicy` in the chart, denying `DELETE` on `etcdmembers` unless the requester is the operator's ServiceAccount, the garbage collector, or the namespace controller — with an explicit break-glass annotation — covers this PR's entire threat list (stray `kubectl delete`, GitOps prune, cleanup scripts) by making the event impossible rather than survivable. Sketch:

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: etcd-operator-protect-members
spec:
  matchConstraints:
    resourceRules:
      - apiGroups: ["etcd-operator.cozystack.io"]
        apiVersions: ["*"]
        operations: ["DELETE"]
        resources: ["etcdmembers"]
  validations:
    - expression: >-
        request.userInfo.username in params.allowedUsers
        || (has(oldObject.metadata.annotations)
            && oldObject.metadata.annotations['etcd-operator.cozystack.io/allow-deletion'] == 'true')
      message: >-
        EtcdMember objects are managed exclusively by etcd-operator, and deleting
        one detaches its data volume permanently. If you really mean it, annotate
        the member with etcd-operator.cozystack.io/allow-deletion=true first.
```

(SA names templated per release; `allowedUsers` = operator SA, `system:serviceaccount:kube-system:generic-garbage-collector`, `system:serviceaccount:kube-system:namespace-controller` so cluster deletion and namespace deletion still cascade. Uninstall docs get one line: remove the policy before removing the CRDs. VAP is GA since Kubernetes 1.30; if we care about older clusters, the same rule fits in a small webhook — but I'd take the version floor.)

Plus the one-line docs fix this PR still doesn't make: `concepts.md` *still* says "Users should not create or edit these directly" — in a PR entirely about deletion semantics, the sentence still doesn't say **delete**, and still doesn't state the consequence.

That pair is small, reversible, has zero migration surface, and closes the hole now. If we still want identity/lifetime separation afterwards as defense-in-depth, I'm open to it — but as its own design discussion, with the blockers below resolved, not as the first line of defense.

## Blockers (if this PR proceeds)

1. **Both member-volume reclaim paths leak the volume permanently after one transient PVC-delete failure.** `controllers/etcdcluster_controller.go` (`scaleDown`) and `controllers/etcdmember_controller.go` (`updateStatus` replacement) both delete the member CR first and the PVC second. The member delete latches instantly; if the PVC delete then fails once, the retry can never reach the reclaim again — the deleting member routes into `handleDeletion` (which never touches the PVC), and on the cluster side `filterRunningMembers` drops the terminating victim so `current > desired` no longer holds and `scaleDown` is never re-entered. The comments call these "the two places the operator reclaims a data volume"; after one apiserver hiccup, neither place ever runs again for that volume. Fix: make reclaim retryable by tying it to the member finalizer — stamp reclaim intent *before* deleting the CR, and in `handleDeletion` delete the PVC after `MemberRemove` succeeds, holding the finalizer on error. Both cases need tests that fail today (interceptor failing the first PVC delete, then asserting the volume is eventually gone).

2. **`handleClusterDeletion` deletes every PVC carrying the cluster label — including volumes the operator never created.** The sweep is scoped only by `MatchingLabels{LabelCluster: name}`. A snapshot-destination PVC (`spec.dest.pvc.claimName` is user-created) that a user labeled with `etcd-operator.cozystack.io/cluster=<name>` — natural, given `docs/operations.md` itself demonstrates that selector for PVCs — is destroyed on cluster deletion. That's the backup being deleted at the exact moment it becomes the only copy. The most destructive loop in the operator must key on the operator's own identity markers, which this very PR introduces: skip any PVC carrying neither `AnnPVCMemberUID` nor a legacy `EtcdMember` owner ref. Related: `deleteMemberPVC` also deletes purely by name with no marker precondition — one `pvcBelongsTo`-style check makes the PR's own invariant ("only volumes the operator created are the operator's to destroy") actually hold everywhere.

3. **Even with (2) fixed, decide explicitly whether cluster deletion may destroy *detached leftovers*.** A volume whose member disappeared out-of-band is precisely the thing this PR promises to protect; the finalizer sweep reclaims it anyway, because it still carries the cluster label and marker. At minimum this needs a loud docs callout here and in the `AllMembersLost` runbook (#343): extract data from detached volumes *before* deleting the cluster. Skipping `detached` volumes whose member no longer exists would be safer still — that's exactly the "operator did not decide this data is finished" case the PR's model is built on.

4. **The upgrade is a one-way door and nothing says so.** `adoptLegacyPVC` migrates every volume on first contact. Rolling back the operator afterwards leaves the old build refusing all of them ("owned by a different EtcdMember; awaiting GC") — management of existing members breaks, and pod recreation after a node event fails. Either gate the migration, or document the irreversibility prominently in release notes and the upgrade docs, including the manual re-add of owner refs as the rollback procedure.

5. **The flagship scenario never gets its `detached` marker.** `handleDeletion` removes the member from etcd and drops the finalizer without touching `AnnPVCStatus`, so after a stray `kubectl delete etcdmember` the surviving volume advertises `ready` forever — contradicting the constant's doc comment, the concepts table, and the PR body ("which is also what a leftover reads"). Stamp `detached` (best-effort) in `handleDeletion` before releasing the finalizer, and extend `TestMemberDeletion_LeavesTheVolumeBehind` to assert it.

6. **Docs regressions.** (a) The new recreate-after-delete recipe waits on `kubectl get etcdcluster,... -l etcd-operator.cozystack.io/cluster=<name>` — but the operator never labels the `EtcdCluster` CR itself, so the terminating cluster is filtered out of the output from the start and the recipe reports "gone" while the finalizer is still draining volumes, inviting exactly the premature re-apply the text warns against; query the cluster by name. (b) `docs/installation.md`'s stranded-operator recovery ("patch finalizers on each `EtcdMember`") no longer unsticks anything: the `EtcdCluster` now holds its own `cluster-cleanup` finalizer and hangs in Terminating with the operator gone — the paragraph needs the cluster patch too, and a note that clearing it forfeits volume reclaim. (c) `cmd/etcd-migrate/apply_adopt.go`'s `adoptPVC` comment still explains the owner ref via `pvcOwnedBy`, which no longer exists — describe the annotation model and the owner ref's new role as a to-be-migrated legacy marker.

## Non-blocking

- `"data-" + member.Name` is still hand-rolled at two call sites in `etcdmember_controller.go`; `memberPVCName()` was added precisely to centralize this.
- Any not-Ready blip downgrades a volume's marker from `ready` back to `initializing`, but the docs define `initializing` as "has not reported Ready *yet*". Consider never downgrading.
- Tiny upgrade window: a legacy PVC can lose its owner ref before the cluster's first post-upgrade reconcile claims `ClusterFinalizer`; deleting the cluster inside that window leaks the volumes. Worth a sentence in upgrade notes.

## Bottom line

Land the admission guard and the docs fix now — small, reversible, closes the hole at the source. Keep the identity/lifetime separation as a follow-up design discussion if we still want it once deletion is guarded; in that discussion, blockers 1–5 are the entry fee, and blocker 3 is the design question that decides whether the model is actually coherent.

One consequence to be explicit about: if the guard direction goes forward, it voids #343 as well. With `EtcdMember` deletion denied at admission, "every member gone while `status.clusterID` is still latched" stops being reachable through any supported path, and the operator would be carrying a terminal condition for a state the platform prevents. I've requested changes there on the same grounds — the two PRs should be decided together.

kvaps:
Accepted — on direction and on every blocker. Implemented your proposal as **#348**: a `ValidatingAdmissionPolicy` denying `DELETE` on `etcdmembers`, plus the docs fix.

Moving this PR back to draft as the follow-up design discussion, not the first line of defence.

## What convinced me

Blocker 2 and point 2 of the direction argument are the same defect seen from two angles, and together they sink this PR as-is: I protected member-level accidents and made cluster-level ones *worse*. The documented recovery after a member-deletion incident is "delete the `EtcdCluster` and recreate" — and my label-scoped `handleClusterDeletion` sweep would erase exactly the volumes the PR had just saved, including detached leftovers that `main` leaves alone. A model whose protection evaporates at the first step of its own runbook is not coherent.

Point 1 lands too. I framed the outcome as "data survives"; what actually survives is bytes that no member can mount, because replacements get fresh names and `pvcBelongsTo` refuses a foreign marker. That is materially better than destroyed bytes, but it is archaeology, not recovery — and the incident still happens.

## Blockers — all verified against the code, all real

1. **Reclaim leaks after one transient failure.** Confirmed both paths. `scaleDown` deletes the CR then the PVC; if the PVC delete fails, the victim is in Terminating, `filterRunningMembers` drops it, `current > desired` no longer holds, and `scaleDown` is never re-entered — while `handleDeletion` never touches the volume. Same shape in the replacement path. Your fix is the right one: stamp intent before deleting the CR and reclaim from the finalizer, holding it on error.
2. **`handleClusterDeletion` deletes by grouping label alone.** Confirmed — and worse than I realised when writing it, since `operations.md` actively teaches that selector for PVCs, so a user-labeled snapshot-destination PVC is a plausible casualty. Must key on `AnnPVCMemberUID` / legacy owner ref; `deleteMemberPVC` needs the same precondition instead of deleting by name.
3. **Detached leftovers vs cluster deletion** — the open design question, agreed. My instinct now matches your safer option: skip volumes whose member no longer exists, since "the operator decided this data is finished" is precisely what did *not* happen for them.
4. **One-way upgrade door.** Correct, and undocumented. `adoptLegacyPVC` strips owner refs fleet-wide on first contact; a downgraded operator then refuses every annotation-only volume via `pvcOwnedBy` and cannot bring pods back after a node event.
5. **Flagship scenario never gets `detached`.** Confirmed — `handleDeletion` releases the finalizer without touching `AnnPVCStatus`, so the surviving volume advertises `ready` forever, contradicting the constant's own doc comment.
6. **Docs regressions** — all three confirmed, including (a): the operator never labels the `EtcdCluster` CR, so my recreate-after-delete recipe filters out the very object it waits on and reports "gone" while the finalizer is still draining.

Non-blocking notes taken as well (the two hand-rolled `"data-" + member.Name` call sites, the `ready` → `initializing` downgrade contradicting the documented meaning, and the upgrade-window leak between owner-ref stripping and the cluster finalizer being claimed).

## If this returns

Then blockers 1–5 first, with blocker 3 settled as a design decision rather than an implementation detail — and the volume-marker precondition applied to *every* deletion path, so the PR's own invariant ("only volumes the operator created are the operator's to destroy") actually holds. I'd also want the CNPG gap you named addressed head-on: there, a leftover instance PVC can be re-attached; here it cannot, and that asymmetry is what turns "survived" into "unmountable forever".

kvaps:
Closing in favour of #348, which implements the admission-guard direction from your review: `DELETE` on `etcdmembers` denied for everyone except the operator, the garbage collector, the namespace controller and an explicit allow-list, with a break-glass annotation — plus the docs fix that finally says **delete** and names the consequence.

The guard is already proving itself in CI: the Kamaji e2e deletes members on purpose and got rejected with the policy's own message, so it now takes the documented break-glass path. That is the accident this branch tried to survive, stopped at the boundary instead.

Nothing here is lost — the analysis, the blockers you found and the CNPG comparison are all on the record in this thread, and the branch stays pushed if the identity/lifetime separation is ever revisited as defence-in-depth. As agreed, that would be its own design discussion, with blockers 1–5 as the entry fee and blocker 3 (may cluster deletion destroy detached leftovers?) settled first, since it decides whether the model is coherent at all.

PR #349, opened 2026-07-30
feat(tls): support handing TLS over from BYO Secrets to operator-managed certs

lllamnyp:
## Problem

A cluster installed with user-provided TLS Secrets has no route onto operator-managed cert-manager issuance. `spec.tls` is CEL-immutable post-create, so the only way to change where the material comes from is delete-and-recreate — which for etcd means destroying the data.

That is a real dead end, not a theoretical one. Charts that predate the v1alpha2 operator issue their own `Certificate`s and hand the Secrets to the cluster via `serverSecretRef` / `secretRef`. Those charts also tend to issue SANs that the operator's `GenerateName`d member names do not match, and there is currently no supported way to move such a cluster onto material the operator generates correctly.

## What this adds

Exactly one permitted post-create change to `spec.tls`:

- `client.serverSecretRef` → `client.certManager`
- `peer.secretRef` → `peer.certManager`

Everything else about the subtree stays frozen. The handover is **one-way**, it may **not** flip whether client mTLS is in effect, and it cannot turn a plane on or off. Adding or removing `tls`, `tls.client` or `tls.peer` is still rejected, as is `certManager` → `secretRef`.

The operator drives the move end to end:

1. **Emit the new Certificates**, refusing to adopt any that another controller already owns.
2. **Wait** until every Secret the new spec names actually carries the keys etcd will be started against.
3. **Repoint every member in one pass**; the member controller rebuilds their Pods.

Progress is reported on a new `TLSHandover` condition: `AwaitingMaterial` → `RollingMembers` → `Complete`, or `Blocked`.

## Three decisions worth reviewing

**The roll is simultaneous, not rolling.** The new material is signed by a different CA, so an old member and a new member cannot authenticate to each other at all — every intermediate state of a staggered roll is a split, quorum-less cluster. A one-at-a-time roll would not avoid the outage, it would stretch it across N pod startups instead of one. So all members are repointed together and come back on consistent material. A brief outage is the accepted cost of the operation.

**Name conflicts fail closed.** The operator's targets are `<cluster>-server` / `<cluster>-peer`. For a cluster named `etcd` — the common case — those are byte-identical to what a chart already emits, and Helm/Flux do not set `ownerReferences` on chart resources, so ownership is a clean discriminator. Adopting the object would mean this operator and the chart's controller reconciling one object forever, and would silently repoint a live cluster at TLS material the operator never issued. Instead it reports:

```
TLSHandover=False  reason=Blocked
  cannot take over Certificate "etcd-peer": it exists but another controller owns it …
```

The unblock is to stop the chart emitting it. cert-manager leaves the Secret behind when its Certificate is deleted, so the operator's replacement Certificate reissues into the same Secret with no gap.

**A conflict does not abort the reconcile.** The cluster is still serving on the material it holds, so the loop continues to `updateStatus` and its health reporting stays live. Bailing out early would freeze `readyMembers` and the health conditions at whatever they were — a cluster that silently stops reporting is worse than one that reports it cannot converge.

## Notable behaviour change

TLS becomes the one mirrored member field that is re-templated after creation. `Storage`, `Resources` and `Version` are deliberately frozen per member — existing members are not re-templated when the cluster spec moves. TLS is different in kind: holding the wrong material does not degrade a member, it isolates it completely. Correspondingly, `ensurePod` now tears down a Pod whose mounted TLS Secret *names* no longer match its member spec (Pod volumes are immutable, so a rebuild is the only route). Content changes — a cert-manager renewal writing a fresh leaf into the same Secret — deliberately do **not** trigger a rebuild.

## Testing

`make manifests generate fmt vet` clean; `go test ./api/... ./controllers/...` green.

New coverage:

- **CEL, via envtest against a real apiserver** (`api/v1alpha2/cel_validation_test.go`): handover accepted on both planes; reverse rejected; a handover that silently drops client mTLS rejected; the mTLS-preserving form accepted; adding a `tls.peer` subtree post-create still rejected.
- **Handover driver** (`controllers/tls_handover_test.go`): no member is touched while material is missing; a Secret that exists but is missing `tls.key` counts as not ready; all members repointed in one pass; each member gets its own copy of the derived TLS view; a conflict is reported without touching members and without taking over the reconcile; no condition appears on a cluster that never drifted; the in-flight condition settles to `Complete`.
- **Conflict guard** (`ensureCertificate`): refuses a foreign-owned Certificate; the pre-existing create-once test was updated to seed an operator-owned Certificate, which is what makes it the "ours, leave it alone" case rather than the conflict case.
- **Pod roll**: a Pod mounting superseded Secrets is deleted; one on current material is left alone (otherwise the member spins in a delete loop).

## Docs

Runbook at `docs/operations.md#handing-tls-over-to-the-operator`, including the condition table and what to do when it stalls. `README.md`, `docs/concepts.md` (validation table and TLS section) and `docs/installation.md` all claimed the subtree was wholly immutable; corrected.

## Summary by CodeRabbit

* **New Features**
  * Added in-place migration from user-provided TLS Secrets to operator-managed cert-manager certificates.
  * Added TLS handover status reporting, including waiting, rolling, blocked, and complete states.
  * Automatically refreshes cluster member Pods when mounted TLS material changes.

* **Bug Fixes**
  * Prevents adoption of certificates managed by another controller.
  * Blocks unsafe or unsupported TLS transitions, including reverse migration.

* **Documentation**
  * Documented TLS handover requirements, limitations, monitoring, and expected brief outage.


lllamnyp:
### On the `kamaji-datastore` e2e failure

The failing test is `TestPVCMemberCrashLoopSelfHeal`, not the Kamaji datastore test — that one passed, including the full member churn and teardown. The failure is pre-existing on `main` and unrelated to this branch:

```
member_selfheal_test.go:93: timed out waiting for stuck member "etcd-8pffl" deleted for
  replacement: victim "etcd-8pffl" still present (crash-loop not yet past threshold)
```

Self-heal is gated on `!member.Spec.Bootstrap`, so when the test happens to pick the bootstrap seed as its victim, replacement can never fire and the wait times out. That is exactly what #345 ("test(e2e): pick a non-seed victim in the self-heal test") fixes; it is still open, and this branch is cut from the current `main` tip, so it inherits the flake.

This change cannot be implicated: the self-heal test builds a deliberately plaintext cluster (`member_selfheal_test.go:54`), so `member.Spec.TLS` is nil and the new Pod-rebuild predicate returns false unconditionally for every member in it.

Happy to rebase once #345 lands if you'd rather see a green run before merging.

### One fix that came out of investigating it

Chasing the above surfaced a real bug in this branch, now fixed in `0e65c23`.

`tlsMountsOutOfDate` keyed off the operator's own volume names (`tls-client` / `tls-peer`). A member adopted in place by `etcd-migrate` keeps the Pod the legacy StatefulSet created — `adoptPod` rewrites labels and owner refs and explicitly does not restart it — so its volumes carry different names while its member spec *does* name TLS Secrets. The predicate read that as drift, and `ensurePod` would have deleted every adopted Pod on the first reconcile after a migration, with no handover requested at all. That is precisely the outcome in-place adoption exists to avoid.

A Pod carrying none of the operator's TLS volumes is now treated as out of scope — it is either plaintext or a shape the operator did not author. Adopted members keep their material until they are rolled onto the operator's own Pod shape, which the runbook now states. Covered by a regression case in `TestTLSMountsOutOfDate` (verified to fail without the guard).

lllamnyp:
Parking this as a draft.

The TLS-reload research (what etcd can and cannot hot-reload) shows the simultaneous roll this PR is built around is both unnecessary and, on memory-backed clusters, unrecoverable:

- etcd re-reads **leaf** cert/key on every handshake, but **CA trust pools are built once at process start and frozen for the process lifetime**. A CA change needs a restart; rewriting the file is silently a no-op.
- etcd trusts multiple CAs concatenated into the single file `--trusted-ca-file` names. So a CA handover decomposes into *widen trust* (one sequential restart per member, no split at any step because every member still presents an old-CA leaf and every member trusts the old CA) and *swap leaves*. There is no intermediate state where a pair of members cannot authenticate, so the roll does not need to be simultaneous.
- `reconcileTLSHandover` has no storage-medium check, and the Pod teardown is a direct delete rather than an eviction, so the emitted PodDisruptionBudget does not gate it. On a memory-backed cluster every member's tmpfs data dir is destroyed in the same instant and no survivor holds the state.

Two further issues in the same area: pointing `--trusted-ca-file` at cert-manager's `ca.crt` narrows the trust set to new-CA-only on reissue — harmless at the time because the pool is frozen, and therefore latent until an unrelated restart isolates a member — and `buildOperatorTLSConfig` takes `ca.crt` from the server Secret, so the operator loses the ability to verify the cluster it is mid-way through converting.

Rather than patch this incrementally, the rollout mechanism is being designed generally: `spec.tls` is only one of several immutable-or-not-honoured settings (storage medium and class, auth, resources, version), and they all want the same thing — a safe, sequenced member roll with a real health gate between steps. This PR will be revisited once that design lands.

The parts of it that stand on their own regardless — the CEL relaxation, the fail-closed conflict guard on foreign-owned Certificates, gating the repoint on material actually being populated, and not rebuilding Pods the operator did not author — will carry over.

Note also that the red `kamaji-datastore` check is unrelated: `TestKamajiDataStore` passes, and the failure is `TestPVCMemberCrashLoopSelfHeal` hitting the seed-victim bug that #345 fixes.

PR #351, opened 2026-07-30
fix(controllers): switch the PDB from maxUnavailable to minAvailable

kjvalencik:
## Production impact

During a node rotation (Karpenter), a healthy 5-voter cluster was drained to 2 voters — below its steady-state quorum of 3 — for ~36 minutes. The budget was `maxUnavailable: 2`, which should have prevented this, but it re-bases under churn: `allowed = currentHealthy - (expectedCount - maxUnavailable)`, where `expectedCount` derives from the /scale subresources of the *currently matching* pods. Each removed member shrank `expectedCount` and refilled the budget mid-drain (evicted-but-terminating voters also briefly kept counting as healthy). No single eviction violated the budget, yet the cluster ended up at 2-of-2.

The window stayed open that long because recovery didn't converge on its own: a replacement pod crashlooped, having been created with a static config that expected a number of voters that had since shrunk.

## Why minAvailable, and why the target anchors it

An integer `minAvailable` has no `expectedCount` term — `allowed = currentHealthy - minAvailable` — so the floor cannot re-base as membership shrinks. The floor is the quorum (`n/2 + 1`) of `max(live voters, status.observed.replicas)`:

- **Steady state**: identical disruptions to the old budget (`n - (n-1)/2 = n/2 + 1`).
- **Unplanned churn**: the latched target holds the floor at steady-state quorum while the operator refills membership. This would have stopped the incident at the first over-budget eviction.
- **Intentional scale-down**: the live count dominates, so the floor tracks the live quorum down as members are removed (via `MemberRemove`, not the eviction API) — a 5→3 shrink never wedges drains.
- **Bootstrap / scale-up**: the target dominates; voter evictions block until the cluster reaches size. Learners are outside the selector and unaffected.

Existing PDBs are migrated in place: the reconciler clears `maxUnavailable` in the same patch that sets `minAvailable` (a PDB with both is invalid). No API change; selector, delete-at-zero-voters, additionalMetadata, and the /scale contract are untouched.

## Future direction

This is deliberately the *least* impactful fix: no new API, steady-state semantics unchanged. What I'd actually like is to allow only a **single** voter to be replaced at a time regardless of cluster size (`minAvailable = n - 1`) — rotations replace nodes serially anyway, and quorum is the emergency floor, not a comfortable operating point. That needs configuration and more complex code, so it's left to a future improvement. Feedback on that design is welcome.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PodDisruptionBudget handling to preserve quorum during disruptions, scaling, and membership changes.
  * Migrated existing budgets from maximum-unavailable rules to minimum-available quorum protection.
  * PodDisruptionBudgets are removed when a cluster has no voting members.
  * Stabilized disruption protection during voter churn and cluster scale transitions.

* **Documentation**
  * Updated user guidance and API descriptions to explain quorum-based disruption protection, scaling behavior, and node-draining expectations.
  * Clarified migration messaging for legacy PodDisruptionBudget configurations.


lllamnyp (approved):

lllamnyp:
@kjvalencik thanks for your work on this, merged.

PR #352, opened 2026-07-31
fix(controllers): extend crash-loop self-heal to memory members

kjvalencik:
The crash-loop self-heal (#336) excluded memory-medium members on the theory that the pod-loss self-heal already covers them. That check only fires when the Pod is gone or replaced (`Status.PodUID` mismatch); a memory member whose Pod is alive but whose etcd can never start keeps the same Pod UID forever and is invisible to it. The operator therefore left such a member crash-looping unrecovered forever.

The live failure mode: a replacement learner is created with `--initial-cluster` baked into its immutable Pod spec; if membership changes again before the learner's first successful boot, etcd fatals with `error validating peerURLs ...: member count is unequal` on every restart. With `max-learners=1` the wedged learner also blocks all further member replacement, so the cluster stays degraded until a human (or a node rotation that happens to kill the Pod) intervenes.

Drop the medium exclusion. Every other guard stays: non-bootstrap, `etcdContainerStuck` (not ready, restart threshold, not OOMKilled, Pod not terminating) and the quorum gate. Replacing a stuck memory member is strictly milder than the already-covered PVC case — its data dies with the Pod anyway.

Includes a regression test (`TestUpdateStatus_ReplacesStuckMemoryMember`) and doc updates in `docs/concepts.md` / `docs/operations.md`.

Related: #351

## Summary by CodeRabbit

* **Bug Fixes**
  * Extended crash-loop self-healing to memory-backed members, enabling automatic replacement when cluster quorum is retained.

* **Documentation**
  * Clarified that crash-loop replacement applies to both memory-backed and PVC-backed members.
  * Documented behavior for stale membership, lost data directories, and memory members without PVC cleanup.


lllamnyp (approved):

lllamnyp:
Hi @kjvalencik , thanks for your amazing work here and on #351. Could you please run `git commit --amend --signoff --no-edit && git push --force-with-lease` on this PR so the DCO check passes?

kjvalencik:
Done, thanks!

Issue #357, opened 2026-08-07
Emit first-party cluster-health metrics and ship a default alert-rule set in the Helm chart


lllamnyp:
The operator currently exposes only stock controller-runtime metrics — the chart ships a ServiceMonitor, and the Cozystack packaging adds a VMServiceScrape, so the scrape surface exists — but no first-party health metrics and no alerting rules. That combination leaves the most damaging failure mode invisible: as #343 established, a cluster can enter a state the reconciler can never leave (e.g. `scaleUp` with zero live endpoints while `status.clusterID` is latched), and the loop retries forever while `status.conditions` keeps reporting `Available=True`. Anything watching conditions sees green; the only truthful signal is the reconcile loop itself.

Proposal, in two halves:

## 1. First-party metrics

Per EtcdCluster, labelled `namespace`/`name` (names are a strawman):

- `etcd_operator_cluster_members_desired` / `_members_ready` / `_members_learners`
- `etcd_operator_cluster_quorum` — 0/1 as currently probed by the operator, not as latched in status
- `etcd_operator_cluster_condition{condition=...}` — terminal/degraded states (`AllMembersLost` from #343, `BootstrapFailed`, …)
- `etcd_operator_cluster_last_successful_reconcile_timestamp_seconds` — the anti-lying heartbeat: consumers alert on its age instead of trusting conditions

## 2. A default alert-rule set shipped by the Helm chart

Values-gated (e.g. `alerts.enabled`), following the pattern cloudnative-pg and piraeus charts already use — the chart ships the rules, the cluster's evaluator just selects them:

- **Reconcile-error streak**: `rate(controller_runtime_reconcile_errors_total{controller=~"etcdcluster.*"}[10m]) > 0` sustained 1h → warning, 6h → critical. Catches every wedged-forever class, including the #343 one, on day one — no new instrumentation needed.
- **Terminal/degraded condition** present → critical immediately.
- `members_ready < members_desired` for 15m → critical.
- `quorum == 0` for 5m → critical.
- `time() - last_successful_reconcile > 1h` per cluster → warning.

The upstream chart already ships prometheus-operator kinds (ServiceMonitor), so PrometheusRule fits there; victoria-metrics-operator conversion makes it consumable in Cozystack deployments, or the Cozystack package can mirror it as a VMRule next to its existing VMServiceScrape.

[...]

Issue #221, opened 2024-05-20
Add defrag CronJob


kvaps:
Kubernetes [recomends](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/#maintaining-etcd-clusters) using etcd-defrag to run defragmentation of the cluster.

Since this is expensive operation this tool allows to specify specific rule, eg:

```
dbQuotaUsage > 0.8 || dbSize - dbSizeInUse > 200*1024*1024
```

https://github.com/ahrtr/etcd-defrag/blob/main/doc/etcd-defrag-cronjob.yaml

This is slightly diferent from kamaji-etcd implementation
https://github.com/clastix/kamaji-etcd/blob/433989a4badeb8edc0925cfc671fb0cc66448e66/charts/kamaji-etcd/templates/etcd_cronjob_defrag.yaml#L11

In Cozystack we currnelty added it into Helm chart:
https://github.com/aenix-io/cozystack/pull/137/files#diff-0dcf3889fea1c88eb22a0d84208697a4ed71c8cb111d1d29ec714a71529d5822

kvaps:
@Kirill-Garbar commented:

> Let's discuss whether it's a cron job or not, just that. Personally, I don't like cron jobs even though they are simple

kvaps:
```yaml
---
apiVersion: etcd.aenix.io/v1alpha1
kind: EtcdCluster
metadata:
  name: test
  namespace: test-tls-auth-etcd-cluster
spec:
  options:
  defrag:
    schedule: "*/15 * * * *"
    rule: # TODO: check the common Kubernetes syntax
     operator: Or
     quotaMoreThan: 80%
     dbSizeMoreThan: 200MB
  storage:
    volumeClaimTemplate:
      spec:
        storageClassName: gp3
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 10Gi
  security:
    enableAuth: true
    tls:
      peerTrustedCASecret: ca-peer-secret
      peerSecret: peer-secret
      serverTrustedCASecret: ca-server-secret
      serverSecret: server-secret
      clientTrustedCASecret: ca-client-secret
      clientSecret: client-secret
 status:
   conditions:
   - lastProbeTime: null
     lastTransitionTime: "2024-03-06T18:39:45Z"
     status: "True"
     type: DefragChecked
```

kvaps:
@lllamnyp also sugested to use go templates:

```
[...]

PR #361, opened 2026-08-16
feat(defrag): EtcdDefrag controller

androndo:
Reworks defragmentation from the rejected `EtcdCluster.spec.defrag` shape (this PR's earlier review) into a **controller for the dedicated `EtcdDefrag` API**.

The `EtcdDefrag` API (#362) has **landed on `main`**; this branch is rebased onto it and targets `main`, so the diff here is the controller + wiring only.

### What it does
Reconciles an `EtcdDefrag` as a one-shot, run-to-completion sweep:
- **Resolve** the referenced cluster; **serialize per cluster** (the oldest non-terminal run acts, the rest wait `Pending`).
- **Real health gate**: every desired member present, reachable, agreeing on a leader, and free of *blocking* alarms — not just "Status answered" (a partitioned/alarmed member still answers a local read). A `NOSPACE` alarm does **not** block: a backend at its quota is exactly the case defrag exists to relieve, so refusing it would leave the one cluster that needs it read-only forever. `CORRUPT` and any other health error still defer.
- **Sweep** members one at a time, **followers before the leader**, one per reconcile pass with health re-checked between. A due defrag on an unhealthy cluster is **deferred** (`DefragChecked=False/ClusterNotHealthy` + a `DefragDeferred` event), never forced.
- **Disarm** the `NOSPACE` alarm once a sweep has reclaimed space, so the cluster leaves read-only (etcd keeps it armed until told otherwise).
- Per-member outcomes/sizes in `status.members`; phase `Pending → Running → Complete|Failed`; an overall active-deadline bounds a stuck run; `ttlSecondsAfterFinished` GCs a finished record.
- **Rule** matches the API: `rule.all` is unconditional; otherwise the reclaimable floor (`freeSpaceAbove`, default 200Mi) is always applied and the quota arm only fires with at least `minReclaim` to reclaim — a full-but-unfragmented backend is never defragmented for nothing.

Adds `Defragment`, `AlarmList`, and `AlarmDisarm` to the etcd client interface; wires the controller (RBAC + `main.go`) and sets `MaxConcurrentReconciles` so a member wedged on one cluster doesn't stall others (per-cluster serialization is enforced separately). **No changes to `EtcdClusterSpec`.**

### Metrics & alerts split out (per review)
The first-party capacity metrics and the values-gated `PrometheusRule` that the earlier version of this PR carried are **removed** — they belong in their own PR against #357 (they must also cover clusters that never opt into defrag, and the metric/label scheme should be settled with the full set in view). This PR ships **no** `controllers/metrics.go`, **no** `charts/.../prometheusrule.yaml`, and **no** `alerts` values; the controller records sizes in `EtcdDefrag.status` during a run instead of as continuously-scraped gauges.

### Deliberately out of scope (follow-ups)
Non-blocking items from the review, left for follow-up PRs:
- Leadership transfer (`MoveLeader`) before defragmenting the leader — leader-last already bounds the ordering risk.
- Re-deriving roles when leadership moves mid-sweep (roles are snapshotted at plan time).
- Raft lag (`RaftIndex - RaftAppliedIndex`) as part of the health gate.
- Graceful handling of a cluster scaled down mid-run (a vanished planned member currently fails the run).
- Restoring end-to-end coverage of the withheld-while-unhealthy → runs-on-recovery property (currently unit-only, see below).

### Tests
- **Unit**: rule evaluation (nil / all / free / quota-with-minReclaim); `markMember` leaves reclaimed/after-size unset when the post-defrag read is unavailable.
- **Controller-integration (fake etcd)**: defragments a fragmented follower (leader last); `rule.all` defragments everyone; **defers with a `DefragDeferred` event and performs no defrag when quorum is lost**; a `NOSPACE` alarm is admitted and disarmed after the sweep; a `CORRUPT` alarm blocks the run; `startedAt` is not re-stamped across a `Pending→Running` flap; failed RPC → `Failed`; two runs on one cluster are serialized.
- **e2e (`//go:build e2e`)**: create an `EtcdDefrag`, fragment a real member and watch its physical `DbSize` shrink to `Complete`. (The withheld-while-unhealthy case is currently unit-only — restoring it to e2e is a follow-up.)

`go build` / `go vet` / `go test ./...` / `gofmt` green; `-tags e2e` compiles; CRD/RBAC/deepcopy regenerated (codegen-drift clean).

Refs #221, #357.

## Summary by CodeRabbit

* **New Features**
  * Added one-shot etcd defragmentation through the `EtcdDefrag` resource.
  * Defragmentation waits for healthy clusters, processes followers before leaders, and reports per-member results and reclaimed space.
  * Added configurable thresholds, deadlines, retries, automatic cleanup, and NOSPACE alarm handling.
  * Added an `install-tools` command for installing required tools.
* **Documentation**
  * Added defragmentation guidance, safety requirements, usage options, and scheduling limitations.
* **Tests**
  * Added comprehensive unit and end-to-end coverage for defragmentation scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ For step-by-step setup, RBAC, image versions, and teardown see [docs/installatio
 - **[Installation](docs/installation.md)** — deploy the operator, create your first cluster, networking pitfalls, upgrades.
 - **[Concepts](docs/concepts.md)** — design rationale: locking pattern, single-seed bootstrap, GenerateName naming, scale-to-zero mechanics, conditions reference.
 - **[Operations](docs/operations.md)** — runbook for day-2: scaling, pausing/resuming, decoding conditions, escalating stuck reconciles, broken-member recovery.
-- **[Defragmentation](docs/etcd-defrag.md)** — the `EtcdDefrag` resource: reclaiming etcd backend disk and its safety model (API type; reconciling controller is a follow-up).
+- **[Defragmentation](docs/etcd-defrag.md)** — the `EtcdDefrag` resource: reclaiming etcd backend disk, one-shot and driven from outside on a schedule, and its safety model.
 - **[Migration](docs/migration.md)** — moving onto this operator from the legacy aenix operator; tracks behavioural changes that need an explicit migration step — currently the BYO root-credentials requirement when enabling auth.
 
 ## Testing
--- a/api/v1alpha2/etcddefrag_types.go
+++ b/api/v1alpha2/etcddefrag_types.go
@@ -38,9 +38,8 @@ type EtcdDefragSpec struct {
 
 	// TTLSecondsAfterFinished records how long after a terminal phase this
 	// object should be garbage-collected — meaningful for objects a scheduler
-	// stamps out. NOTE: acted on by the (not-yet-implemented) reconciling
-	// controller; the API server does not garbage-collect custom resources on
-	// its own. Absent means the record is kept.
+	// stamps out. Acted on by the reconciling controller; the API server does not
+	// garbage-collect custom resources on its own. Absent means the record is kept.
 	// +kubebuilder:validation:Minimum=0
 	// +optional
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
--- a/api/v1alpha2/etcddefrag_types.go
+++ b/api/v1alpha2/etcddefrag_types.go
@@ -70,9 +70,10 @@ type DefragRule struct {
 	// QuotaUsageAbove: when DbSize exceeds this fraction of the backend quota
 	// (approaching NOSPACE), lower the reclaimable floor to MinReclaim so small
 	// wins are taken under pressure. A member is never defragmented when its
-	// reclaimable space is below MinReclaim. Integer percent 1..100 with a "%"
-	// suffix, e.g. "80%".
-	// +kubebuilder:validation:Pattern=`^([1-9][0-9]?|100)%$`
+	// reclaimable space is below MinReclaim. Integer percent 1..99 with a "%"
+	// suffix, e.g. "80%"; 100% is rejected because a backend never exceeds its
+	// quota (etcd raises NOSPACE first), so the arm could never fire.
+	// +kubebuilder:validation:Pattern=`^[1-9][0-9]?%$`
 	// +optional
 	QuotaUsageAbove string `json:"quotaUsageAbove,omitempty"`
 
--- a/api/v1alpha2/etcddefrag_types.go
+++ b/api/v1alpha2/etcddefrag_types.go
@@ -201,10 +201,6 @@ type MemberDefragStatus struct {
 // run-to-completion defragmentation of an EtcdCluster's members. Like
 // EtcdSnapshot it is a record: the operator drives it through status.phase and
 // it never re-runs.
-//
-// NOTE: this ships the API type ahead of its reconciling controller. Until that
-// controller lands, an EtcdDefrag is inert — creating one records intent but
-// nothing acts on it (no sweep runs, status stays empty, TTL does not fire).
 type EtcdDefrag struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml
+++ b/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml
@@ -132,9 +132,8 @@ spec:
                 description: |-
                   TTLSecondsAfterFinished records how long after a terminal phase this
                   object should be garbage-collected — meaningful for objects a scheduler
-                  stamps out. NOTE: acted on by the (not-yet-implemented) reconciling
-                  controller; the API server does not garbage-collect custom resources on
-                  its own. Absent means the record is kept.
+                  stamps out. Acted on by the reconciling controller; the API server does not
+                  garbage-collect custom resources on its own. Absent means the record is kept.
                 format: int32
                 minimum: 0
                 type: integer
--- a/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml
+++ b/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml
@@ -35,10 +35,6 @@ spec:
           run-to-completion defragmentation of an EtcdCluster's members. Like
           EtcdSnapshot it is a record: the operator drives it through status.phase and
           it never re-runs.
-
-          NOTE: this ships the API type ahead of its reconciling controller. Until that
-          controller lands, an EtcdDefrag is inert — creating one records intent but
-          nothing acts on it (no sweep runs, status stays empty, TTL does not fire).
         properties:
           apiVersion:
             description: |-
--- a/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml
+++ b/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml
@@ -110,9 +110,10 @@ spec:
                       QuotaUsageAbove: when DbSize exceeds this fraction of the backend quota
                       (approaching NOSPACE), lower the reclaimable floor to MinReclaim so small
                       wins are taken under pressure. A member is never defragmented when its
-                      reclaimable space is below MinReclaim. Integer percent 1..100 with a "%"
-                      suffix, e.g. "80%".
-                    pattern: ^([1-9][0-9]?|100)%$
+                      reclaimable space is below MinReclaim. Integer percent 1..99 with a "%"
+                      suffix, e.g. "80%"; 100% is rejected because a backend never exceeds its
+                      quota (etcd raises NOSPACE first), so the arm could never fire.
+                    pattern: ^[1-9][0-9]?%$
                     type: string
                 type: object
                 x-kubernetes-validations:
--- a/charts/etcd-operator/files/manager-role-rules.yaml
+++ b/charts/etcd-operator/files/manager-role-rules.yaml
@@ -1,6 +1,13 @@
 - apiGroups:
     - ""
   resources:
+    - events
+  verbs:
+    - create
+    - patch
+- apiGroups:
+    - ""
+  resources:
     - persistentvolumeclaims
   verbs:
     - create
@@ -87,12 +94,24 @@
     - etcd-operator.cozystack.io
   resources:
     - etcdclusters/status
+    - etcddefrags/status
     - etcdmembers/status
     - etcdsnapshots/status
   verbs:
     - get
     - patch
     - update
+- apiGroups:
+    - etcd-operator.cozystack.io
+  resources:
+    - etcddefrags
+  verbs:
+    - delete
+    - get
+    - list
+    - patch
+    - update
+    - watch
 - apiGroups:
     - etcd-operator.cozystack.io
   resources:
--- a/controllers/etcd_client.go
+++ b/controllers/etcd_client.go
@@ -50,6 +50,15 @@ type EtcdClusterClient interface {
 	// its embedded Maintenance interface.
 	Defragment(ctx context.Context, endpoint string) (*clientv3.DefragmentResponse, error)
 
+	// AlarmList and AlarmDisarm let the EtcdDefrag controller close the NOSPACE
+	// loop: a cluster at its backend quota raises NOSPACE and goes read-only,
+	// which is the case defrag exists to relieve, so the health gate permits the
+	// run — but the alarm stays armed after the space is reclaimed until it is
+	// explicitly disarmed. A CORRUPT alarm, by contrast, blocks the run.
+	// *clientv3.Client satisfies both via its embedded Maintenance interface.
+	AlarmList(ctx context.Context) (*clientv3.AlarmResponse, error)
+	AlarmDisarm(ctx context.Context, m *clientv3.AlarmMember) (*clientv3.AlarmResponse, error)
+
 	// Auth surface — used by reconcileAuth to provision the single root
 	// user/role and turn on authentication. The "root" role is built into
 	// etcd, so a RoleAdd is not needed: UserAdd("root", …) +
--- a/controllers/etcd_client.go
+++ b/controllers/etcd_client.go
@@ -35,12 +35,20 @@ type EtcdClusterClient interface {
 	MemberPromote(ctx context.Context, id uint64) (*clientv3.MemberPromoteResponse, error)
 	MemberRemove(ctx context.Context, id uint64) (*clientv3.MemberRemoveResponse, error)
 
-	// Status returns a single endpoint's server status, including the etcd
-	// version it is actually running (StatusResponse.Version). Used by the
-	// member controller to observe the running version into
-	// EtcdMember.status.version. *clientv3.Client satisfies this via its
+	// Status returns a single endpoint's server status: the etcd version it is
+	// actually running (StatusResponse.Version, observed into
+	// EtcdMember.status.version), its backend sizes (DbSize / DbSizeInUse), the
+	// leader it sees and any alarms — all read by the EtcdDefrag controller to
+	// gate and decide a defragmentation. *clientv3.Client satisfies this via its
 	// embedded Maintenance interface.
 	Status(ctx context.Context, endpoint string) (*clientv3.StatusResponse, error)
+
+	// Defragment releases a single endpoint's reclaimable backend space to the
+	// filesystem. It is per-endpoint (defrag is a member-local operation) and
+	// briefly blocks that member, so the EtcdDefrag controller calls it one
+	// member at a time on a healthy cluster. *clientv3.Client satisfies this via
+	// its embedded Maintenance interface.
+	Defragment(ctx context.Context, endpoint string) (*clientv3.DefragmentResponse, error)
 
 	// Auth surface — used by reconcileAuth to provision the single root
 	// user/role and turn on authentication. The "root" role is built into
--- a/controllers/etcd_client.go
+++ b/controllers/etcd_client.go
@@ -59,6 +59,18 @@ type EtcdClusterClient interface {
 	AlarmList(ctx context.Context) (*clientv3.AlarmResponse, error)
 	AlarmDisarm(ctx context.Context, m *clientv3.AlarmMember) (*clientv3.AlarmResponse, error)
 
+	// MoveLeader asks the member serving this call to hand raft leadership to
+	// transfereeID. The EtcdDefrag controller calls it before defragmenting the
+	// leader, so the stop-the-world pause lands on a follower instead of costing
+	// an election.
+	//
+	// Unlike Status/Defragment this is NOT per-endpoint: clientv3 sends it to
+	// whichever endpoint its balancer picks, and etcd answers ErrNotLeader
+	// anywhere but the leader. Callers must therefore invoke it on a client
+	// dialled with the leader as its only endpoint — see dialEndpoints.
+	// *clientv3.Client satisfies this via its embedded Maintenance interface.
+	MoveLeader(ctx context.Context, transfereeID uint64) (*clientv3.MoveLeaderResponse, error)
+
 	// Auth surface — used by reconcileAuth to provision the single root
 	// user/role and turn on authentication. The "root" role is built into
 	// etcd, so a RoleAdd is not needed: UserAdd("root", …) +
--- a/controllers/etcddefrag_controller.go
+++ b/controllers/etcddefrag_controller.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
+	etcdserverpb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -27,6 +28,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	lll "github.com/cozystack/etcd-operator/api/v1alpha2"
@@ -53,6 +55,10 @@ const (
 	// so a run stuck on an unhealthy cluster can't hold the per-cluster slot
 	// forever.
 	defragActiveDeadline = 30 * time.Minute
+
+	// defragMaxConcurrentReconciles lets distinct clusters' runs proceed in
+	// parallel (per-cluster serialization is enforced separately by oldestActive).
+	defragMaxConcurrentReconciles = 4
 )
 
 // EtcdDefragReconciler drives an EtcdDefrag: a one-shot, run-to-completion
@@ -141,13 +147,20 @@ func (r *EtcdDefragReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	defer c.Close()
 
 	// Health gate: defrag is only safe when every desired member is present,
-	// reachable, alarm-free and agrees on a leader.
+	// reachable, free of blocking alarms and agreeing on a leader.
 	if reason, ok := clusterDefragHealthy(cluster, running, backends); !ok {
 		msg := "a defragmentation is due but the cluster is not fully healthy; deferring to protect quorum: " + reason
 		if setDefragCondition(df, metav1.ConditionFalse, "ClusterNotHealthy", msg) {
 			r.event(df, corev1.EventTypeWarning, "DefragDeferred", msg)
 		}
-		df.Status.Phase = lll.EtcdDefragPhasePending
+		// Keep a run that has already started (partial results in
+		// status.members) in Running and let the condition carry the reason;
+		// only a run that never started falls back to Pending. StartedAt is what
+		// the active-deadline is measured against, so it must not be re-stamped
+		// by such a flap — see the Running transition below.
+		if df.Status.StartedAt == nil {
+			df.Status.Phase = lll.EtcdDefragPhasePending
+		}
 		if err := r.Status().Update(ctx, df); err != nil {
 			return ctrl.Result{}, err
 		}
@@ -161,7 +174,7 @@ func (r *EtcdDefragReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 	next := firstPendingMember(df)
 	if next == nil {
-		return r.finalize(ctx, df)
+		return r.finalize(ctx, df, c)
 	}
 
 	b := backendByName(backends, next.Name)
@@ -171,12 +184,19 @@ func (r *EtcdDefragReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return r.persistAndRequeue(ctx, df)
 	}
 
-	if df.Status.Phase != lll.EtcdDefragPhaseRunning {
-		df.Status.Phase = lll.EtcdDefragPhaseRunning
+	df.Status.Phase = lll.EtcdDefragPhaseRunning
+	// Stamp StartedAt exactly once, on the first Running transition. The
+	// active-deadline is measured from it, so re-stamping on a Pending->Running
+	// flap (a cluster that recovers between health-gate blips) would keep
+	// resetting the deadline and a stuck run could hold the per-cluster slot
+	// forever.
+	if df.Status.StartedAt == nil {
 		now := metav1.Now()
 		df.Status.StartedAt = &now
-		setDefragCondition(df, metav1.ConditionTrue, "Running", "defragmenting members")
 	}
+	// Refresh the condition each active pass so a run that resumes after a
+	// health-gate flap does not stay reading ClusterNotHealthy.
+	setDefragCondition(df, metav1.ConditionTrue, "Running", "defragmenting members")
 
 	quota := effectiveQuotaBytes(cluster)
 	if trig, reason := defragRuleTriggered(df.Spec.Rule, b.status.DbSize, b.status.DbSizeInUse, quota); !trig {
@@ -194,22 +214,32 @@ func (r *EtcdDefragReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return r.persistAndRequeue(ctx, df)
 	}
 
-	// Read the post-defrag size for the record (best-effort).
+	// Read the post-defrag size for the record. When the read fails, leave the
+	// after-size unset rather than defaulting it to the before-size — reporting
+	// a successful defrag as reclaiming zero would be silently wrong in the very
+	// field the per-member status exists to provide.
+	outcomeReason := ""
 	if after, serr := statusWithTimeout(ctx, c, b.endpoint); serr == nil {
 		b.after = after.DbSize
+		b.afterKnown = true
+	} else {
+		outcomeReason = "AfterSizeUnavailable"
+		logger.Info("defrag: post-defrag Status read failed; reclaimed size unrecorded",
+			"member", b.member.Name, "err", serr)
 	}
-	markMember(df, b.member.Name, lll.DefragOutcomeDefragmented, "", b)
+	markMember(df, b.member.Name, lll.DefragOutcomeDefragmented, outcomeReason, b)
 	df.Status.Defragmented++
-	logger.Info("defragmented member", "member", b.member.Name, "reclaimed", b.status.DbSize-b.after)
+	logger.Info("defragmented member", "member", b.member.Name)
 	return r.persistAndRequeue(ctx, df)
 }
 
 // memberBackend pairs a running member with its live Maintenance Status.
 type memberBackend struct {
-	member   *lll.EtcdMember
-	endpoint string
-	status   *clientv3.StatusResponse
-	after    int64 // DbSize after a successful defrag
+	member     *lll.EtcdMember
+	endpoint   string
+	status     *clientv3.StatusResponse
+	after      int64 // DbSize after a successful defrag; valid only if afterKnown
+	afterKnown bool  // whether the post-defrag Status read succeeded
 }
 
 // dialAndProbe opens one client and reads every running member's Status. The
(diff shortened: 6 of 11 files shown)


lllamnyp (changes requested):
## Requesting changes — this belongs outside `EtcdClusterSpec`

First, credit where it's due: the safety model here is the right one. One member per pass, followers before the leader, refusing to stamp the cooldown after a failure, deferring rather than forcing on a degraded cluster, and documentation that is honest about its own approximations rather than overselling them. The reasoning in "Design vs alternatives" for hosting this in the operator — TLS/auth material and endpoints already in hand, whole-cluster view already available, leader election already free — is correct, and I don't want that argument thrown away.

What I can't take is the shape. Defragmentation is operational intent that runs on a completely different clock from `EtcdCluster` reconciliation. Reconciling an `EtcdCluster` is level-triggered convergence toward a declared target: it should be idempotent, cheap, and finish. Defragmentation is an occasional, expensive, stateful, ordered procedure with its own cadence, its own failure modes, and its own history. Folding the second into the first buys nothing and costs the clarity of both.

The API consequence is the part I care most about, because it's the permanent part. `spec.defrag` sets a precedent: every operational concern gets a block in `EtcdClusterSpec`. Next it's compaction policy, then snapshot scheduling, then rebalancing, then whatever comes after — each with its own thresholds, cadence, and rule sub-object, all of them mutually independent and all of them living in the spec of a resource whose job is to describe *what the cluster is*, not *what maintenance we'd like performed on it*. An operator exists to take operational burden off the user. A spec that grows a knob per procedure hands it back with extra steps.

### Two ways forward

**Option A — ship the CronJob, as #221 was filed.** A per-cluster `CronJob` running `etcd-defrag` with the rule from the issue. Genuinely less capable: it re-plumbs TLS/auth/endpoints per cluster, it's blind to whether the operator is mid-scale-up, and it has no cluster-wide view. But it's small, it's what the issue asked for, and it adds zero permanent API. If defragmentation isn't worth a resource of its own, it isn't worth `spec.defrag` either — this is the honest floor.

**Option B — give it its own resource.** An `EtcdDefrag` with a `.spec.clusterRef`, reconciled by its own controller:

- The intent becomes a first-class object with its own lifecycle. Its `status` is the natural home for what's currently being crammed into a cluster condition and a member annotation — per-member progress, last-run times, outcomes, the whole ordered sweep. You get history and observability for free, and you can express "defrag this cluster once, now" as a one-shot object, which is a thing operators actually want at 3am and which `spec.defrag` cannot express at all.
- The reconciler runs on its own clock and its own work queue, and stops interfering with the cluster reconciler's schedule and algorithm. Right now the two are entangled in a way that's already causing bugs (below).
- `EtcdClusterSpec` stays a description of the cluster. Snapshot scheduling, compaction policy, and whatever comes next follow the same pattern instead of each landing as another spec block.

If you take Option B, one more thing is worth trying, since most of the work is already done: **drive the trigger from metrics rather than from a rule embedded in the CRD.** Thresholds belong where thresholds already live — in alerting rules, evaluated by the metrics stack, tuned without a CRD change and without an operator release. An `EtcdDefrag` that reacts to a custom/external metrics source (the way an HPA does) rather than carrying `quotaUsageAbove` / `freeSpaceAbove` in its own spec keeps the API to "defrag this cluster when this signal says so" and puts the polling burden on the system built for polling. It also removes the need for the operator to probe every member's `Maintenance.Status` on its own loop just to decide whether to act. Treat this as optional and gated on a fallback for clusters with no metrics adapter — but if the effort is being spent anyway, it's the version that ages best.

### The coupling is already costing you

Three defects in this PR are not really independent bugs; they're the same structural decision surfacing three times. `reconcileDefrag` returns early from `Reconcile` (`controllers/etcdcluster_controller.go:383-391`) on paths that pre-empt `updateStatus`, and `updateStatus` is where this operator maintains `Available`, `Degraded`, `readyMembers`, `brokenMembers`, and the PodDisruptionBudget.

1. **An unparseable `spec.defrag.schedule` freezes the cluster's entire status and its PDB.** `controllers/defrag.go:85-93` returns with no requeue, so `updateStatus` never runs and the 30s heartbeat is gone with it. A cluster with a typo'd cron keeps serving while `Available` reads whatever it last said — including `True/QuorumHealthy` after a member has died.

2. **The blocked-defrag path starves `updateStatus` for the whole degraded window.** `controllers/defrag.go:187-195` returns with a 15s requeue whenever a defrag is due and the cluster isn't fully healthy. For a PVC-backed member replacement — tens of minutes, per the README — health conditions and the PDB stop being updated in exactly the window where they matter, while the operator re-dials etcd and re-probes every member every 15 seconds against an already-degraded cluster.

3. **`DefragChecked` is a write-once latch, not a live signal.** `controllers/defrag.go:180-182` sets the condition in memory and falls through, but `updateStatus` writes status only when *its own* comparisons changed — it has no idea a condition was mutated upstream. Reproduce by settling every field `updateStatus` touches and running a not-needed pass: the condition is never persisted at all on a healthy under-threshold cluster, and after a successful run it stays `True/Defragmented` indefinitely. `DefragNotNeeded` is documented as a reachable steady state; in practice it usually isn't. (`TestReconcileDefrag_NotNeeded` misses this because it asserts against the in-memory object where its sibling tests re-`Get`.)

Under Option B all three dissolve, because a separate reconciler owns a separate status object. That's the argument for the restructure in concrete form.

### Findings that survive either path

These are etcd-domain issues, independent of where the code lives — worth carrying forward into whichever option you take:

- **The health gate can't detect the failure it exists to prevent** (`controllers/defrag.go:129-146`). `Maintenance.Status` is a local read: a member answers it while having lost quorum, while carrying a `NOSPACE` or `CORRUPT` alarm, and while arbitrarily far behind in raft. A partitioned cluster with all pods up and reachable passes the gate. The response already carries `Errors`, `Leader`, `RaftIndex`, `RaftAppliedIndex`, and `IsLearner`; none are checked. At minimum require `Leader != 0` with agreement across members, and an empty `Errors`.
- **No leadership transfer before defragging the leader** (`controllers/defrag.go:197-208`). "Leader last" bounds the ordering risk, not the leader-specific one: a defrag longer than the election timeout costs an election. `MoveLeader` is already on the `clientv3` Maintenance interface.
- **The quota arm re-triggers forever with nothing to reclaim** (`controllers/defrag.go:248-271`). `dbSize == dbSizeInUse == 1.7Gi` against a 2Gi quota returns true under both the default rule and an explicit `quotaUsageAbove: 80%`. With the 1h default cooldown that's an hourly stop-the-world of every member, reclaiming nothing, forever, with a `Defragmented` event each time. Needs a minimum-reclaimable gate on the quota arm, and a backoff when a defrag doesn't shrink `DbSize`.
- **Nothing compacts, and nothing disarms the alarm.** Auto-compaction is an unset-by-default user knob, so a cluster without it has `dbSizeInUse ≈ dbSize`: the free-space arm never fires, the quota arm fires forever, and defrag reclaims nothing. And a cluster that actually reaches `NOSPACE` stays read-only after a successful defrag, because the alarm is still armed — `kubectl etcd alarm disarm` already exists in this repo, but the operator never notices the alarm and never clears it. Either close that loop or stop implying in the README and runbook that this recovers a quota-pressured cluster.
- **The shipped alerts break under the ServiceMonitor the same chart ships.** `charts/etcd-operator/templates/prometheusrule.yaml:20-22,32-34` join `on(namespace, cluster)`, but prometheus-operator relabels every target's `namespace` from the scrape target and `honor_labels` defaults to false, so the metric's own `namespace` becomes `exported_namespace` and `namespace` becomes the operator's. Two clusters sharing a name in different namespaces then make the right-hand side ambiguous and both quota rules stop evaluating — in a multi-tenant install that's the common case, not the edge case. Needs `honorLabels: true` on the endpoint in `servicemonitor.yaml` (plus the equivalent for the Cozystack `VMServiceScrape`), or a different label name.
- **The capacity gauges go missing for seconds on every pass** (`controllers/defrag.go:125-127`, re-set at `:143-144`). The `DeletePartialMatch` runs before a probe loop that spends up to 5s per unreachable member, so the series are absent for up to N×5s each pass. Any scrape landing in that gap resets the `for: 15m` timer — the slow-cluster case the alert exists for may never fire. Build the new value set first, then delete only the labels that dropped out.
- **`EtcdDefragmentationNotKeepingUp` false-positives against its own defaults** (`prometheusrule.yaml:45-47`): it fires at the same 200Mi that is the default trigger, with `for: 30m` shorter than the default `minInterval: 1h`. A member that re-fragments and waits out its cooldown trips a warning while working as designed.
- **No admission validation on the new numeric fields** (`api/v1alpha2/etcdcluster_types.go:339-370`). `minInterval` accepts `0s` and negatives, which removes the cooldown entirely and lets defrag run on every pass; `freeSpaceAbove` accepts `0` and negatives through the stock Quantity pattern, making the arm always-true. `quotaUsageAbove` is properly pattern-validated — match it.
- **`schedule` doesn't schedule** (`controllers/defrag.go:291-307`). The first run ignores it entirely, and afterwards the firing time drifts to whatever hour the threshold happens to be crossed. The godoc is honest about this, but a cron-shaped field named `schedule` will be read as "run the disruptive thing at 3am" by everyone who sets it and the docs won't be re-read. Either implement a real window or name the field for what it does. A separate resource makes the real thing easy.

### Split the metrics out

The capacity metrics and the `PrometheusRule` should be their own PR against #357, not a passenger on this one. They're the smaller, lower-risk half; they're currently emitted only inside `reconcileDefrag`, so a cluster that hasn't opted into defrag — the one with no automatic remediation and the most need for a quota alert — exports nothing at all; and the gauges linger at stale values forever if a user later removes `spec.defrag`, since they're only cleared on cluster deletion. They're also spending #357's metric namespace and label scheme on one slice of what that issue asked for, before the rest of it is designed: #357's own strawman labels these `namespace`/`name`, this ships `namespace`/`cluster`, and `cluster` additionally collides with the external label Thanos/Mimir/VictoriaMetrics add in multi-cluster fleets. Metric names and labels are as permanent as API — worth settling with the full set in view.

### What I'd approve

Either option, done properly. Option A is a small PR and I'll take it on its merits. For Option B: an `EtcdDefrag` type with `.spec.clusterRef`, its own controller and status, no new fields in `EtcdClusterSpec`, the domain findings above addressed, and the metrics split into a separate PR. The metrics-driven trigger is a bonus, not a condition.

Happy to talk through the `EtcdDefrag` shape before you invest in it — the API is the part worth agreeing on first, and the mechanism you've already written should port across largely intact.

lllamnyp (changes requested):
Structurally this is what the split was for. The cluster reconciler is untouched, so the defrag path can no longer pre-empt `updateStatus` or stall PDB reconciliation; the health gate checks `status.Errors`, a non-zero `Leader`, and leader agreement across members instead of trusting a bare local read; runs are serialized per cluster; there is a per-RPC bound and an overall run deadline; and `status.members[]` carries real per-member records. The `minReclaim` floor does what it was meant to — a full-but-unfragmented backend is no longer defragmented for nothing.

Two defects need fixing before this lands. Both reproduce against the existing fake-etcd test harness.

## Blocking

### A `NOSPACE` alarm blocks defragmentation entirely

`clusterDefragHealthy` (`controllers/etcddefrag_controller.go:274`) fails the gate on any non-empty `status.Errors`, and that field is where etcd reports active alarms — the API defines it as "errors contains alarm/health information and status".

On a 3-member cluster with `rule.all: true` and one member alarmed:

```
phase=Pending defragCalls=[]
condition DefragChecked=False reason=ClusterNotHealthy
  msg=... not fully healthy ...: member c1-0 reports alarms: memberID:10 alarm:NOSPACE
```

No `Defragment` call is ever issued. The run sits in `Pending` until `defragActiveDeadline` fails it 30 minutes later.

A cluster that has hit its backend quota is the case this feature exists to resolve, and it is the one cluster the controller now refuses to touch. The gate needs to discriminate between alarms rather than blanket-refusing: `NOSPACE` should permit the run, `CORRUPT` should block it. Closing that loop properly also means disarming the alarm after a successful sweep — `AlarmList`/`AlarmDisarm` are not yet on the `EtcdClusterClient` interface, and without them a cluster stays read-only after the space has been reclaimed.

### `status.startedAt` is reset on every `Pending → Running` transition, so the run deadline never expires

Lines 174-179 stamp `StartedAt` whenever the phase is not already `Running`, and the health-gate branch at line 150 sets the phase back to `Pending` on every blip. Seed a run whose clock already reads 25 of its 30 minutes, park it in `Pending`, and give it one healthy pass:

```
seeded startedAt 25m ago; after one healthy pass it is 0s old (phase=Running)
```

The comment on `defragActiveDeadline` (lines 52-55) states the guarantee this is supposed to provide — that a run stuck on an unhealthy cluster cannot hold the per-cluster slot forever. On a cluster that flaps more often than every 30 minutes, the run never fails and every subsequent `EtcdDefrag` for that cluster queues behind it indefinitely, including ones a scheduler stamps out. Fix is to stamp only when `StartedAt == nil`.

## Non-blocking

**No `MoveLeader` before defragmenting the leader.** Doing the leader last bounds the ordering risk but not the leader-specific one: a defrag that outlasts the election timeout costs an election and a write-availability blip. `MoveLeader` is already on the `clientv3` Maintenance interface.

**Roles are snapshotted once and never re-derived.** `plannedMembers` fixes each member's role at plan time, so "followers before the leader" holds only against the plan-time leader. Leadership can move mid-sweep — defragmenting a member can itself cause it — after which the new leader is processed as a follower while the old one waits at the end of the list. Re-deriving the leader each pass, or reordering when the current leader comes up next while followers remain pending, would close it.

**`MaxConcurrentReconciles` is unset (so 1) while `defragRPCTimeout` is 5 minutes**, so one wedged member stalls defragmentation for every other cluster too. Since `oldestActive` already enforces per-cluster serialization, this is safe to raise.

**A failed post-defrag `Status` read reports a successful defrag as reclaiming nothing.** `b.after` is pre-seeded to the pre-defrag `DbSize` at probe time (line 244), so when the read at line 198 fails, `DBSizeAfter` equals `DBSizeBefore` and `ReclaimedBytes` is 0. That is silently wrong in the field the per-member status design exists to provide; better to leave the after-size unset and say the read was unavailable.

**Smaller things:**

- The phase moves backwards `Running → Pending` on a health blip after members have already been processed. A run holding partial results reading `Pending` is confusing; the condition already carries the reason, so the phase could stay `Running`.
- Raft lag is not part of the health gate. A member that has rejoined and is still catching up answers `Status`, reports no alarms, and agrees on the leader, so blocking a second member with a defrag can still stall writes. `RaftIndex - RaftAppliedIndex` is in the same response already being read.
- `docs/etcd-defrag.md:133` states that "a defrag that doesn't shrink `DbSize` is backed off rather than repeated". The controller does not do this, and inside a one-shot run that touches each member once there is nothing for it to mean. Either drop the line or move it to whatever ends up owning repeat scheduling.
- A cluster that legitimately scales down mid-run leaves a planned member absent, which marks it `MemberGone`/`Failed` and fails the whole run.

## Test coverage

The six unit tests cover the sweep, `rule.all`, deferral on an unhealthy cluster, a failed RPC, and per-cluster serialization, which is a good spread. Not covered: TTL garbage collection, deadline expiry, and leadership drift mid-sweep — the second of which is where one of the blockers above lives.

The e2e suite went from two tests to one; the case that proved a due defragmentation is withheld while the cluster is unhealthy and runs once it recovers no longer has end-to-end coverage, leaving the safety property on the unit test alone.

## Merge order

This is stacked on #362, which is good to go. Land #362 first; GitHub will retarget this to `main` automatically.

lllamnyp (changes requested):
Reviewed against `main` at merge base `a371f0c`, which already carries the `EtcdDefrag` API from #362.

Green here: `go build ./...`, `go vet ./controllers/... ./api/...`, `go vet -tags e2e ./test/e2e/...`, `go test ./controllers -run 'TestDefrag|TestEtcdDefrag|TestMarkMember'`, `gofmt -l`, and `make manifests generate` leaves the tree clean. I also confirmed empirically the alarm-string contract the health gate depends on: `(&etcdserverpb.AlarmMember{MemberID: 10, Alarm: AlarmType_NOSPACE}).String()` really is `"memberID:10 alarm:NOSPACE "` on etcd 3.6.11, and `a.String()` is what `maintenance.Status` appends to `resp.Errors` across the 3.5/3.6 releases — so `blockingStatusError`'s substring match is sound.

The blockers below are the NOSPACE disarm path — the headline behaviour of this revision — and a set of documentation claims that the shipped controller does not honour.

---

## Blocking

### 1. A partially-failed sweep never disarms NOSPACE, so the cluster stays read-only forever

`controllers/etcddefrag_controller.go:517-533`

```go
phase := lll.EtcdDefragPhaseComplete
for _, m := range df.Status.Members {
    if m.Outcome == lll.DefragOutcomeFailed {
        phase = lll.EtcdDefragPhaseFailed
        ...
    }
}
if phase == lll.EtcdDefragPhaseComplete && df.Status.Defragmented > 0 {
    if err := r.disarmNoSpaceAlarms(ctx, c); err != nil { ... }
}
```

**What's wrong.** The disarm is gated on the run being *entirely* clean. On a 3-member NOSPACE cluster where two members defragment successfully and the third's `Defragment` RPC fails (or the member vanished → `MemberGone`, line 183), the run goes `Failed` and `disarmNoSpaceAlarms` is never called — even though gigabytes were actually reclaimed on the two that succeeded. etcd keeps the alarm armed until told otherwise, so the cluster stays read-only. And because terminal phases are sticky, this `EtcdDefrag` will never retry; the user must create a new one, which will find little left to reclaim (the successful members are already compact) and is likely to fail the same way on the same sick member. The one situation this feature exists to rescue — a quota-wedged, read-only cluster — is precisely the situation where the disarm is skipped.

The same hole opens on the deadline path: `fail(..., "DeadlineExceeded", ...)` at lines 128-132 goes straight to `Failed` without ever touching the alarm, so a sweep that reclaimed space on 4 of 5 members and then hit the 30-minute wall leaves the cluster read-only too.

**Fix.** Decouple "did we reclaim space" from "was the run clean". Disarm whenever `df.Status.Defragmented > 0`, regardless of the final phase, and do it on the `fail()` path as well (or, better, immediately after the first successful member defrag, since etcd re-arms on the next write if the space was not actually freed — that is what makes the call safe to make eagerly). Concretely, change the guard to `if df.Status.Defragmented > 0` in `finalize`, and give `fail` access to the client so a deadline-terminated run that reclaimed something still disarms.

**Test that fails without the fix.** In `controllers/etcddefrag_controller_test.go`, a variant of `TestEtcdDefrag_NoSpaceAlarmPermitsRunAndDisarms` with `rule.all` on a 3-member cluster, `fe.alarms = [{MemberID:10, Alarm: NOSPACE}]`, all three endpoints reporting a NOSPACE line, and the fake failing `Defragment` for exactly one endpoint (extend `fakeEtcd` with a `defragErrByEndpoint map[string]error` alongside the existing `defragErr`). Drive to terminal and assert `len(fe.disarmCalls) == 1`. Today it is 0 and the phase is `Failed`.

### 2. `disarmNoSpaceAlarms` aborts on the first failure, leaving the remaining members armed

`controllers/etcddefrag_controller.go:297-308`

```go
for _, a := range resp.Alarms {
    if a == nil || a.Alarm != etcdserverpb.AlarmType_NOSPACE { continue }
    ...
    if derr != nil {
        return derr        // <- abandons every later alarm
    }
}
```

**What's wrong.** `AlarmList` returns one `AlarmMember` per member that raised NOSPACE. A transient failure disarming the first one abandons the loop, so members 2..N stay armed and the cluster stays read-only. Worse, the caller (`finalize`, line 530-532) only *logs* the error, so the run still reports `Complete` — the status says the defrag succeeded while the cluster is still refusing writes. The comment above the function calls this "best-effort", but the loop is not best-effort: it is fail-fast with a swallowed error.

**Fix.** Continue the loop on error, accumulate with `errors.Join`, and return the joined error so the single log line names every member that could not be disarmed. Additionally, surface a non-nil result to the user rather than only the log — e.g. a `Warning`/`AlarmDisarmFailed` event via `r.event`, so `kubectl describe etcddefrag` shows why the cluster is still read-only after a `Complete` run.

**Test that fails without the fix.** Add `disarmErrByMember map[uint64]error` to `fakeEtcd.AlarmDisarm`. Seed `fe.alarms` with three NOSPACE members (10, 11, 12), fail the disarm for member 10, run `rule.all` to `Complete`, and assert `len(fe.disarmCalls) == 3`. Today it is 1.

### 3. Docs claim raft lag is part of the health gate; it is not

`docs/etcd-defrag.md:105-107`

> Health is judged from more than "the member answered": a member replies to a local status read while partitioned, alarmed (`NOSPACE`/`CORRUPT`), or **behind in raft, so those are checked before acting**.

**What's wrong.** `clusterDefragHealthy` (`controllers/etcddefrag_controller.go:320-347`) checks member count, reachability, blocking alarms and leader agreement. It never reads `RaftIndex`/`RaftAppliedIndex` — the PR description itself lists raft lag as a deliberate follow-up. The sentence promises a safety property the code does not provide, on the exact page an operator reads before trusting this with their quorum.

The same sentence is doubly wrong: it lists `NOSPACE` among the alarms that are "checked before acting", while the whole point of this revision is that a NOSPACE alarm is deliberately *admitted* (`blockingStatusError`, lines 354-362). A reader of the Safety model section is told the opposite of what ships.

**Fix.** Either implement the raft-lag check in `clusterDefragHealthy` (compare `status.RaftIndex - status.RaftAppliedIndex` against a constant, deferring above it) or rewrite the bullet to state what is actually checked: every desired member present and reachable, no `CORRUPT` (or other non-`NOSPACE`) alarm, and unanimous agreement on a non-zero leader — with `NOSPACE` explicitly called out as admitted rather than blocking, and raft lag named as not yet covered.

**Test that fails without the fix.** If you implement the check: a `clusterDefragHealthy` unit case with three reachable members agreeing on a leader where one reports `RaftIndex: 1000, RaftAppliedIndex: 100`, asserting `ok == false`. Today it returns `true`. If you instead fix the prose, the guard is issue #2's / #4's e2e-and-unit coverage plus a docs read — do not add a grep-the-markdown test.

### 4. Docs claim per-member RPC retries and leader-only run failure; neither exists

`docs/etcd-defrag.md:130-132`

> **Retry within a run:** a deferred `Pending` re-checks cluster health **with backoff** up to the deadline; a failed per-member RPC **is retried a bounded number of times** then marked `Failed` (**a failing leader fails the run**).

Three claims, three mismatches against `controllers/etcddefrag_controller.go`:

- **No retry.** Lines 209-215: the *first* `Defragment` error calls `markMember(..., DefragOutcomeFailed, "RPCError", b)`. There is no attempt counter anywhere in the file; the member is terminal after one failure.
- **No backoff.** Line 167 (and 121, 566) returns a fixed `RequeueAfter: defragRequeueAfter` = 10s. The deferral interval is constant, not backed off.
- **Any member fails the run, not just the leader.** `finalize`, lines 520-526, scans all of `status.members` and flips to `Failed` on the first `Failed` outcome regardless of role. The parenthetical tells operators a failed follower is survivable; it is not, and — per issue #1 — that difference is what leaves a NOSPACE cluster read-only.

**Fix.** Preferably implement the bounded per-member retry the doc promises (a `retries` counter on `MemberDefragStatus`, or a bounded requeue before marking `Failed`), since a single transient gRPC error on one follower currently kills the whole run and, on a NOSPACE cluster, leaves it wedged. If retries stay out of scope, rewrite the bullet to say a failed per-member RPC marks that member `Failed` immediately and that any failed member fails the run, and drop "with backoff".

**Test that fails without the fix.** If retries are implemented: extend `fakeEtcd` so `Defragment` fails the first N calls for an endpoint then succeeds, and assert the run reaches `Complete` with that member `Defragmented` and `len(fe.defragCalls) > 3` for a 3-member `rule.all`. Today the run is `Failed` after a single call to that endpoint.

### 5. The `EtcdDefrag` CRD still tells users the controller does not exist

`api/v1alpha2/etcddefrag_types.go:39-46`, generated into `charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_etcddefrags.yaml:132-137`

```go
// TTLSecondsAfterFinished records how long after a terminal phase this
// object should be garbage-collected — meaningful for objects a scheduler
// stamps out. NOTE: acted on by the (not-yet-implemented) reconciling
// controller; the API server does not garbage-collect custom resources on
// its own. Absent means the record is kept.
```

**What's wrong.** This PR carefully strips the "inert until the controller lands" notes from the `EtcdDefrag` type doc, the CRD description, `docs/etcd-defrag.md` and the README — and misses this one. It is not a code comment: `controller-gen` bakes it into the CRD's OpenAPI schema, so `kubectl explain etcddefrag.spec.ttlSecondsAfterFinished` tells every operator on the cluster that the field is inert, when `handleTTL` (`controllers/etcddefrag_controller.go:571-583`) now implements it.

**Fix.** Drop "(not-yet-implemented)" from the Go doc comment — e.g. "acted on by the reconciling controller; the API server does not garbage-collect custom resources on its own" — and re-run `make manifests` so the CRD picks it up. (I verified `make manifests generate` is otherwise drift-free, so this is the only regeneration needed.)

**Test that fails without the fix.** None — the guard here is issue #6's real TTL test plus the CI codegen-drift gate, which already re-renders the CRD from the marker. Per the repo's own convention, do not add a test that greps the CRD YAML for the string.

### 6. `ttlSecondsAfterFinished` is newly live and completely untested

`controllers/etcddefrag_controller.go:571-583`

**What's wrong.** This PR is what makes `spec.ttlSecondsAfterFinished` do anything, `docs/etcd-defrag.md:45-47` advertises it ("with `ttlSecondsAfterFinished` the controller GCs the record an hour after it finishes"), and the RBAC diff adds `delete` on `etcddefrags` solely to support it — but `controllers/etcddefrag_controller_test.go` never exercises `handleTTL`. Nothing catches a regression in the expiry arithmetic, the requeue-until-expiry branch, or the delete itself. Two live paths are silently broken-able:

- `handleTTL` returns `ctrl.Result{}, nil` with no requeue when `Status.CompletedAt == nil`, so a terminal object without a completion stamp is never GC'd and never re-examined.
- The requeue for a not-yet-expired object is `expiry.Sub(now)`, which is only re-evaluated if a reconcile actually happens — an operator restart between `finalize` and expiry is handled by the startup resync, but nothing proves it.

The same gap covers two other newly-added terminal paths with zero tests: `DeadlineExceeded` (lines 127-133) and `ClusterNotFound` (lines 102-105). `etcdsnapshot_controller_test.go` has tests for both of its equivalents (`TestSnapshotReconcile_JobActiveDeadlineExceededFails`, and the `ClusterNotFound` assertion at line 68), so this controller is below the bar its sibling already sets.

**Fix.** Add the three tests.

**Tests.**
- TTL: create a `Complete` `EtcdDefrag` with `TTLSecondsAfterFinished: ptrInt32(3600)` and `Status.CompletedAt` stamped 2h ago; reconcile once; assert `client.IgnoreNotFound(c.Get(...))` reports the object gone. Then the same with `CompletedAt` stamped 1s ago and assert the object survives with `0 < res.RequeueAfter <= 3600s`.
- Deadline: seed `Status.StartedAt` 31 minutes ago with the cluster unreachable (`fe.statusErr` set); reconcile once; assert `Phase == Failed` and the `DefragChecked` condition reason is `DeadlineExceeded`.
- ClusterNotFound: an `EtcdDefrag` whose `spec.clusterRef.name` matches no `EtcdCluster`; assert `Phase == Failed`, reason `ClusterNotFound`, and a `Warning` event.

### 7. The run deadline is not related to the worst-case sweep it has to bound

`controllers/etcddefrag_controller.go:49-57`

```go
defragStatusTimeout = 5 * time.Second
defragRPCTimeout    = 5 * time.Minute
defragRequeueAfter  = 10 * time.Second
defragActiveDeadline = 30 * time.Minute
```

**What's wrong.** These four constants are independent literals, but the sweep is strictly serial: one member per reconcile pass, each pass costing up to `N × defragStatusTimeout` for the probe loop in `dialAndProbe` plus up to `defragRPCTimeout` for the defrag plus `defragRequeueAfter` before the next pass. For a 7-member cluster (legal, and the size most likely to have a large backend) the worst case is `7 × (5m + 10s + 35s) ≈ 40m` against a 30-minute deadline: a run where *every member is behaving and slow* is killed mid-sweep by `DeadlineExceeded`. Combined with issue #1, a big-backend cluster under NOSPACE gets a half-completed defrag, a `Failed` record, and an alarm that is never disarmed.

Note the reviewer-facing symptom: this is not a hypothetical timeout tuning nit, it is the interaction that decides whether the feature works on exactly the clusters it was written for (large, fragmented backends — the only ones where a 5-minute defrag RPC is realistic).

**Fix.** Derive the deadline from the plan instead of hardcoding it: compute it once when the plan is built as `len(df.Status.Members) × (defragRPCTimeout + defragRequeueAfter) + slack`, and store it (or at least assert the relationship with a package-level constant both sides reference so the compiler fails closed if someone edits one). Alternatively raise `defragActiveDeadline` and add a comment stating the maximum member count it covers.

**Test that fails without the fix.** A unit test asserting the deadline covers the worst-case serial sweep for the largest supported cluster: `require defragActiveDeadline >= 7 * (defragRPCTimeout + defragRequeueAfter + 7*defragStatusTimeout)`. Today `30m < ~40m` and it fails. (This is a compile-adjacent constant relation, not a source-grep — it reads the constants the controller actually uses.)

---

## Recommended to fix (non-blocking)

- **`docs/etcd-defrag.md:13-16`** — "A companion `EtcdDefragPolicy` kind … is planned so the operator absorbs that scheduling itself; **it is not part of this API PR**." Shipped documentation should not reference "this API PR"; a reader on the website has no PR in view, and after this branch merges the page is no longer an API-only PR at all. Reword to "not implemented yet" and link the tracking issue.

- **`controllers/etcddefrag_controller.go:180-185` (`MemberGone`)** — this branch marks the member `Failed` and persists, but never sets `df.Status.Phase = Running` and never stamps `StartedAt` (both happen later, at lines 187-196). A run whose *first* planned member vanishes therefore stays in `Pending` while already carrying a `Failed` member row, and its deadline keeps being measured from `CreationTimestamp` rather than from the start of work. Move the phase/`StartedAt` transition above the `backendByName` lookup.

- **`controllers/etcddefrag_controller.go:206-233` (status-update staleness)** — `df` is read at the top of the pass and written after a `Defragment` that may block for `defragRPCTimeout` (5 min). Any unrelated write to the object in that window (a `kubectl label`, a `kubectl annotate`) makes `r.Status().Update` fail with a conflict, discarding the recorded outcome and the `Defragmented++`; the member is still `Pending` next pass and gets defragmented a second time. Harmless in effect but wasteful and confusing in the record. Re-fetch and re-apply the member row on conflict (`retry.RetryOnConflict`) rather than returning the error.

- **`controllers/etcddefrag_controller.go:471` / `parsePercent`** — `quotaUsageAbove: "100%"` passes the CRD pattern but is dead: the arm requires `float64(dbSize) > 1.0 * float64(quota)`, and etcd raises NOSPACE and refuses writes before the backend can exceed its quota. Either tighten the CRD pattern to 1..99 or document that 100% never fires.

- **NOSPACE + default rule** — on a quota-wedged cluster whose members are full but *unfragmented* (`DbSize ≈ DbSizeInUse`), every member is `Skipped`, `Defragmented == 0`, no alarm is disarmed, and the run reports `Complete`. That is the correct action (there is nothing to reclaim), but the user gets a green `Complete` on a still-read-only cluster with no explanation. Consider a distinct terminal reason (e.g. `NothingToReclaim`) and an event when the run completes with `Defragmented == 0` while a NOSPACE alarm is armed, so `kubectl describe` points at compaction rather than at the operator.

---

## What is solid

Worth saying explicitly, because these were the risky parts and they hold up:

- The alarm-string contract in `blockingStatusError` matches real etcd output on both 3.5 and 3.6 (verified empirically, above) — including the `NOSPACE`-skip / `CORRUPT`-block asymmetry and the `etcdserver: no leader` line.
- The health gate checks leader *agreement*, not just "Status answered", which is the distinction that actually protects quorum.
- `StartedAt` is stamped exactly once and the `Pending→Running` flap test pins it — that is the bug that would otherwise let a stuck run hold the per-cluster slot indefinitely.
- Per-cluster serialization via `oldestActive` is deterministic (creationTimestamp with a name tiebreak), so two workers racing on two `EtcdDefrag`s for one cluster reach the same answer without a lock.
- `markMember` leaving `dbSizeAfter`/`reclaimedBytes` unset when the post-defrag read fails, rather than reporting a real defrag as reclaiming zero, is the right call and is tested.
- Codegen, RBAC and the e2e build tag are all clean; the e2e test fragments a real backend and asserts a physical `DbSize` shrink rather than trusting the controller's own status.

---

lllamnyp (approved):
Approving. Every blocker from the previous round is closed, and I checked each against the code rather than the commit messages.

**Disclosure: the last commit on this branch is mine.** `fix(defrag): move leadership off the leader before defragmenting it` was the one gap I was still holding this on, and rather than ask for another round I pushed it here. I can't meaningfully review my own change — @sircthulhu, if you look at one commit, make it that one.

### Resolved

- A `NOSPACE` alarm no longer blocks the run, and a partial or deadline-terminated sweep that reclaimed space still disarms on the way out. On a 3-member cluster with one member alarmed and a later member's RPC failing, the run ends `Failed` with 2 members defragmented and the alarm disarmed — the wedge is lifted even though the run did not finish cleanly.
- `disarmNoSpaceAlarms` continues past a per-member failure, joins the errors naming each, and surfaces `AlarmDisarmFailed`, so a still-read-only cluster after a run is visible in `kubectl describe`.
- `status.startedAt` is stamped once, so the active deadline no longer restarts on a health flap and a stuck run cannot hold the per-cluster slot indefinitely.
- The deadline is derived from the constants that bound the worst-case serial sweep — 45m15s against a 40m15s worst case — with a relation test so retuning any input keeps them consistent.
- The docs no longer claim raft lag is gated or that per-member RPCs are retried, and they state the shipped behaviour instead: a failed RPC fails the member and the run, no backoff.
- The stale "not-yet-implemented" note is gone from the CRD, so `kubectl explain` no longer tells operators a live field is inert.
- TTL garbage collection, deadline expiry, and cluster-not-found now have tests; the suite went from 10 to 20 cases.

### Non-blocking, for whoever picks up the follow-ups

- **Roles are snapshotted once** in `plannedMembers` and never re-derived, so "followers before the leader" holds against the plan-time leader. Leadership can still move mid-sweep for reasons other than the transfer above.
- **Raft lag is not part of the health gate.** A member that has rejoined and is still catching up answers a status read, reports no alarm and agrees on the leader, so blocking a second member with a defrag can still stall writes. `RaftIndex`/`RaftAppliedIndex` are already in the response being read. Correctly documented as not-yet-gated.
- **The conflict retry is only on `persistAndRequeue`.** The terminal writes in `fail` and `finalize` still use a bare status update, and `failRun` runs the disarm before its write, so the same window exists there — smaller, but the treatment is inconsistent.
- **Tightening `quotaUsageAbove` to 1..99 narrows validation on a CRD that already merged.** No stored object plausibly carries `100%` this early, so the practical risk is nil, but it is worth being a deliberate call. The justification is also marginally stronger than the truth: etcd rejects writes once size plus cost reaches the quota, but bbolt allocates in pages, so `dbSize` can edge past it. `100%` is very nearly always inert rather than provably always — still the right call, a threshold that only fires in a pathological corner is a footgun.
- **e2e is still a single test.** The case proving a due defragmentation is withheld while the cluster is unhealthy and runs once it recovers has no end-to-end coverage; the unit coverage for it is good, so this is a gap in the real-cluster tier only.

PR #362, opened 2026-08-17
feat(api): add EtcdDefrag API type and docs

androndo:
Adds a dedicated **`EtcdDefrag`** resource for operator-driven backend defragmentation — the Option B shape from the review on #361 (which accepted the defrag *mechanism* but rejected hosting it as `EtcdClusterSpec.defrag`).

This PR lands the **API type + generated deepcopy/CRD + user docs** only, so the reconciling controller can follow as a self-contained change with no API churn at that point. Until the controller lands the resource is inert (documented as such). **No changes to `EtcdClusterSpec`.**

### What's here
- `api/v1alpha2/etcddefrag_types.go` — `EtcdDefrag` (+`List`), `EtcdDefragSpec{clusterRef, rule?, ttlSecondsAfterFinished?}`, `DefragRule`, `EtcdDefragStatus` with per-member `status.members[]`, phase `Pending→Running→Deferred→Complete|Failed`. Registered in the scheme.
- Generated `zz_generated.deepcopy.go` and the `etcddefrags` CRD (chart auto-globs `crd-bases/`, so it renders with no manual wiring — `helm template` shows 4 CRDs, `helm lint` clean).
- [`docs/etcd-defrag.md`](docs/etcd-defrag.md) + a README link.

### Shape, briefly
- **One-shot, modeled on `EtcdSnapshot`.** Recurring is driven externally (a `CronJob` creating objects), matching the repo's deliberate lack of `EtcdSnapshotSchedule`. This removes fields that were defect sources in #361: no `schedule` (cadence is the scheduler's), no `minInterval`.
- **`rule` bakes in a reclaimable floor** (`freeSpaceAbove`, and `minReclaim` under `quotaUsageAbove` pressure) so a full-but-unfragmented backend isn't defragmented forever for nothing — the correctness fix expressed in the API, with CEL validation on the numeric fields.
- **`status.members[]`** holds per-member role/outcome/before-after size/reclaimed — history/observability as first-class status, not a condition + annotation.

### Deliberately out of scope
- The **controller** (follow-up; the mechanism from #361 ports across largely intact) — its safety contract (real health gate on `Leader`/`Errors`/agreement, `MoveLeader` before the leader, `NOSPACE` disarm, per-cluster serialization, no-op backoff) is written up in the doc so the API is judged against real semantics.
- **Capacity metrics + alert rules** → their own PR against #357, with the full label scheme in view.

Refs #221, #357; supersedes the `spec.defrag` approach in #361.

## Summary by CodeRabbit

* **New Features**
  * Added the `EtcdDefrag` resource for one-time etcd defragmentation requests.
  * Supports cluster selection, reclaim-space and quota-pressure rules, lifecycle status, per-member outcomes, validation, printer columns, and automatic cleanup settings.
  * The resource currently records intent only; defragmentation is not yet processed automatically.

* **Documentation**
  * Added guidance on scheduling, safety behavior, retries, timeouts, capacity metrics, and current limitations.
  * Linked the defragmentation documentation from the README.


lllamnyp (changes requested):
This is the right shape, and it answers what #361 was missing: a dedicated resource with `.spec.clusterRef`, the controller deferred so the API can be judged on its own, and `EtcdClusterSpec` left alone. `status.members[]` — per-member role, outcome, before/after `DbSize`, bytes reclaimed, as a `listType=map` — is a real improvement on a rolled-up condition plus an annotation, and gives a run genuine history. The `rule` redesign also fixes the worst correctness defect of the earlier approach *in the API itself*: making reclaimable space the always-applied floor, with `minReclaim` as the floor under quota pressure, is what stops a full-but-unfragmented backend from being defragmented hourly forever for nothing.

Codegen is clean (`make generate manifests` produces no drift, so the deepcopy and CRD are genuinely tool-generated), `go build ./...` passes, and `helm template` renders 4 CRDs.

Two things to change: the scheduling stance, and a validation bug.

## Scheduling: the resource is right, the doc's conclusion isn't

`docs/etcd-defrag.md:16-17` states that there is deliberately no `schedule` field because cadence is an external scheduler's job. That forecloses a decision that should go the other way.

Defragmentation needs to be expressible three ways — once, on a cadence, and when a condition is met. All three produce the same thing: a sweep across members with per-member outcomes and a terminal result. They differ only in what starts the sweep. So the split is between *the run* and *what triggers a run* — the Job/CronJob decomposition:

| What the user wants | How they say it |
|---|---|
| Once, now | Create an `EtcdDefrag` by hand |
| On a cadence | `EtcdDefragPolicy` with `schedule` |
| When a condition is met | `EtcdDefragPolicy` with `when` |
| On a cadence, but only if it's worth doing | `EtcdDefragPolicy` with **both** |

The last row is why this shape wins. Schedule and condition are not three modes, they are two independent predicates that compose, and the intersection — "at 03:00, and only if there is something to reclaim" — is what a careful operator actually wants. Neither predicate alone expresses it, and a `oneOf` trigger union inside a single kind would forbid it by construction.

```yaml
apiVersion: etcd-operator.cozystack.io/v1alpha2
kind: EtcdDefragPolicy
spec:
  clusterRef: {name: etcd}

  schedule: "0 3 * * *"      # optional: when we are allowed to act
  when:                       # optional: whether it is worth acting
    freeSpaceAbove: 200Mi
    quotaUsageAbove: 80%
  minInterval: 6h             # floor between runs; only meaningful for `when`-only

  suspend: false
  concurrencyPolicy: Forbid   # per-cluster serialization, in vocabulary users already know
  template:                   # the EtcdDefrag that gets stamped out
    spec:
      rule: {freeSpaceAbove: 200Mi, quotaUsageAbove: 80%, minReclaim: 32Mi}
      ttlSecondsAfterFinished: 3600
status:
  lastScheduleTime: ...
  lastSuccessfulTime: ...
  active: {name: etcd-defrag-28f3a}
```

`schedule` only means cron. `when` only means level-triggered, rate-limited by `minInterval`. Both means the schedule gates when the operator may look and `when` decides whether to bother. Neither should be rejected at admission (`has(self.schedule) || has(self.when)`), so a policy that can never fire cannot be created.

**`when` and `rule` are not redundant, and the docs will need to say why.** `when` decides whether to create a run at all — it is what stops a level-triggered policy from stamping out a no-op `EtcdDefrag` on every reconcile. `rule` decides which members a run actually touches, evaluated at execution time against live sizes. Different layers, different clocks: a run created at 03:00 may find the fragmentation already gone, and `rule` is what makes that a cheap no-op instead of three pointless stop-the-world calls.

Why two kinds rather than one with a `trigger` union: a single kind forces one status to serve two lifecycles. `phase: Complete`, `completedAt`, and `ttlSecondsAfterFinished` are meaningful for a one-shot and meaningless for a standing object; `lastScheduleTime` and `active` are the reverse. Half the status inapplicable depending on spec is the shape to avoid.

This also does not reintroduce what #361 was rejected for. That objection was to accumulating operational concerns inside `EtcdClusterSpec`. Two small single-purpose kinds with clean lifecycles is the opposite — and it removes the external `CronJob` that the current doc requires, which is burden the operator should be absorbing rather than handing back.

On naming: prefer `EtcdDefragPolicy` over `EtcdDefragSchedule`. It holds a condition as well as a cadence, so `Policy` is the honest name, and it avoids echoing the `EtcdSnapshotSchedule` the README rules out.

**What this changes in this PR: structurally, nothing.** `EtcdDefrag` as proposed is already the run type in all four rows above — `clusterRef`, `rule`, `ttlSecondsAfterFinished`, terminal phases, and `status.members[]` make it a clean Job-analog. What changes is:

- `docs/etcd-defrag.md:16-17` should say a policy kind follows, not that external-only scheduling is deliberate and permanent.
- The owner-reference question below stops being optional: policy-created runs need one, both for garbage collection and so `concurrencyPolicy` can find the active run.
- `README.md:79` currently advertises the doc as covering defragmentation "one-shot and scheduled", while the README's own feature list still says "No defragmentation scheduling". That becomes true rather than aspirational once the policy kind exists; until then, drop "and scheduled".

## Blocking

### Both new CEL rules reject integer-form quantities with a raw type error

`api/v1alpha2/etcddefrag_types.go:50-51` call `quantity(self.freeSpaceAbove)` and `quantity(self.minReclaim)` directly. Every other `quantity()` call in this repo coerces first — `quantity(string(self.storage.size))` in `etcdcluster_types.go:376` and `:429`. Because these are `int-or-string` fields, integer input matches no overload. Verified against a real apiserver with the repo's envtest harness on Kubernetes 1.33:

| Input | Result |
|---|---|
| `freeSpaceAbove: "200Mi"` | accepted |
| `freeSpaceAbove: 200` | **rejected** — `'no such overload: quantity(int)'` |
| `freeSpaceAbove: 0` | **rejected** with the same overload error, not the intended message |
| `minReclaim: 32` | **rejected** — `'no such overload: quantity(int)'` |

A plain byte count is a legitimate `Quantity`, so `freeSpaceAbove: 209715200` fails with an incomprehensible CEL error that has `"freeSpaceAbove must be greater than 0"` glued onto the end of it.

This exact path already has a named regression test on `EtcdCluster` — `TestCEL_StorageMustBeNonZero_IntegerInput` in `api/v1alpha2/cel_validation_test.go` — whose comment spells out that `quantity()` requires a string and that omitting `string()` trips a "no such overload" runtime error instead of the intended human-readable message. Fix is `quantity(string(...))` in both rules, plus the equivalent integer-input test for `EtcdDefrag`.

### Ship order: this releases an inert CRD and a doc written in the present tense

The chart auto-globs `crd-bases/`, so the next chart release installs `etcddefrags` with nothing reconciling it — someone creates one during an incident and gets silence. `docs/etcd-defrag.md:3-5` carries a status banner, but the sections that follow ("Safety model", "Timeouts and retries") describe guarantees in the present tense as though they exist.

Either stack the controller and land them together, or: move the inert warning into the Go type's godoc so it reaches `kubectl explain`, and mark those two sections as planned behaviour.

## Non-blocking, worth fixing

**`rule: {}` and an absent `rule` mean opposite things.** Per the godoc, omitting `rule` means "defragment every member unconditionally" (`etcddefrag_types.go:31-33`), while `rule: {}` gets the 200Mi default gate (`:55`). Both are accepted. So the present-but-empty object is the safe form and omission is the aggressive one — the inverse of the usual reading, and a trap for any chart or GitOps template that emits `rule: {}` when its values block is empty. Consider an explicit `rule.all: true`, or requiring `rule` outright, so "unconditional" is something the user says rather than something they get by leaving a key out.

**`spec.clusterRef` with an empty name is accepted.** Both `clusterRef: {}` and `clusterRef.name: ""` pass admission: `required: [clusterRef]` only requires the object, and `corev1.LocalObjectReference.Name` carries `default: ""`. This is inherited rather than invented — `EtcdSnapshot.spec.clusterRef` has the identical weakness — but this CRD is being defined right now, so a `self.clusterRef.name != ''` rule is free here, where changing `EtcdSnapshot` later is a separate conversation.

**Two `rule` combinations are accepted but incoherent.** `minReclaim` with no `quotaUsageAbove` is a silent no-op: it floors an arm that was never enabled. And `minReclaim: 1Gi` alongside `freeSpaceAbove: 200Mi` is accepted, which inverts the field's purpose — under quota pressure the gate becomes stricter than normal. One CEL rule each.

**Smaller things:**

- `Outcome`, `Role`, and `Reason` (`etcddefrag_types.go:129-142`) are bare strings whose valid values live only in prose, while `Phase` gets typed constants. Define them now so the follow-up controller and any consumer share one source.
- `Deferred` as a phase overlaps `Pending` — both mean "not acting yet" — and forces every consumer to know it is non-terminal. It may read better as a condition on a `Pending` object.
- `ttlSecondsAfterFinished` is well named and well motivated for scheduler-created objects, but `EtcdSnapshot` has no equivalent. If it is right here it is probably right there; the two record types now diverge.
- Nothing specifies what happens to `EtcdDefrag` records when their `EtcdCluster` is deleted. An owner reference would collect them; TTL will not, since it only fires after a terminal phase and a `Pending` orphan never reaches one.

## Coordination

`DefragRule` is declared in `api/v1alpha2` by both this PR and #361, so they cannot both merge. Since this supersedes that approach, #361 should be closed rather than left open.

lllamnyp (approved):
Everything raised earlier is addressed: `quantity(string(...))` in both rules with an integer-input test alongside the existing `EtcdCluster` one, a non-empty `clusterRef.name`, `rule.all` replacing the absent-versus-empty inversion (with CEL keeping it exclusive of the thresholds), `minReclaim` gated on `quotaUsageAbove` and bounded by `freeSpaceAbove`, typed `DefragOutcome`/`MemberRole`, `Deferred` demoted from a phase to a condition, and the inert-until-the-controller-lands warning where `kubectl explain` will show it.

The scheduling section now reads correctly: `EtcdDefrag` is the run, a companion `EtcdDefragPolicy` carrying a cadence and/or a condition is planned to trigger runs, and neither is foreclosed by what ships here.

Codegen is drift-free and `helm template` renders the CRD set. Land this ahead of #361, which stacks on it and removes the inert warnings as the controller arrives.

Issue #186, opened 2024-04-21
Antipattern in status checks


lllamnyp:
# Summary

The status field of the `EtcdCluster` is self-referential and is not determined by the status of an actual etcd cluster.

# Background

Kubernetes controlelrs are [level based](https://book-v1.book.kubebuilder.io/basics/what_is_a_controller.html). This generally means, that a controller should be able to determine its status from the state of the surrounding environment. It is generally considered a bad idea to read the status of the root object (in this case, the `EtcdCluster` CR).

> Remember, status should be able to be reconstituted from the state of the world, so it’s generally not a good idea to read from the status of the root object. Instead, you should reconstruct it every run.
>
> Kubebuilder book — [Implementing a controller](https://book.kubebuilder.io/cronjob-tutorial/controller-implementation#2-list-all-active-jobs-and-update-the-status)

Controllers also do not distinguish between types of events, which trigger reconciliation (was a resource created, updated, deleted, etc). Trying to fight this restriction may cause complications further down the line. For example, implementing statuses, which describe the lifecycle of a resource and then relying on them to select a mode of operation of the controller can lead to new failure modes which need to be handled in different ways, growing the complexity of the codebase.

# Issue

In its current state, the controller relies on an `EtcdCluster`'s status field to determine the next steps in the reconciliation logic. If the controller reads an incorrect status without actually validating it, the etcd cluster will fail to launch. A contrived example can produce this issue as follows:

<details>
<summary>First apply the following yaml</summary>

```yaml
---
apiVersion: etcd.aenix.io/v1alpha1
kind: EtcdCluster
metadata:
  name: test
  namespace: default
spec:
  replicas: 3
  podTemplate:
    spec:
      containers:
      - name: etcd
        image: "nginx"
        args: ["-c", "sleep 3600"]
        command: ["/bin/sh"]
        livenessProbe:
          httpGet:
            host: ifconfig.me
            path: /
            port: 80
            scheme: HTTP
        readinessProbe:
          httpGet:
            host: ifconfig.me
            path: /
            port: 80
            scheme: HTTP
        startupProbe:
          httpGet:
            host: ifconfig.me
            path: /
            port: 80
            scheme: HTTP
```

</details>

This object will generate a statefulset which will launch and pass all readiness checks, but will not in fact build a working cluster, instead three containers will simply run `sleep 3600`. The cluster-state configmap will set the value of `ETCD_INITIAL_CLUSTER_STATE` to `existing`.

<details>
[...]

PR #369, opened 2026-09-15
fix(controllers): keep EtcdCluster status truthful when etcd is unreachable

### LINKED ISSUE #367: EtcdCluster status freezes at its last good value when etcd is unreachable
## Summary

When the operator cannot reach an EtcdCluster's etcd endpoints, the cluster reconciler returns early and never reaches `updateStatus`. `status.readyMembers`, `Available`, `Degraded` and `Progressing` keep whatever values they had the last time etcd answered. A cluster whose every member is down keeps reporting `Available=True / QuorumHealthy / "All members are ready"` and `readyMembers: 3` indefinitely, while its EtcdMembers all report `Ready=False`.

Observed on a Cozystack integration cluster running etcd-operator v0.5.4 after a graceful node shutdown took every etcd pod down at once. `main` (a3d1b19) has the same code.

## What it looks like

```
$ kubectl get etcdclusters -A
NAMESPACE         NAME   VERSION   REPLICAS   READY   AGE
tenant-portal     etcd   3.6.11    3          3       5d7h

$ kubectl get etcdmembers -A
NAMESPACE         NAME         CLUSTER   VERSION   RUNNING   READY   AGE
tenant-portal     etcd-7p9fr   etcd      3.6.11    3.6.11    False   5d7h
tenant-portal     etcd-9wm4c   etcd      3.6.11    3.6.11    False   5d7h
tenant-portal     etcd-smx9c   etcd      3.6.11    3.6.11    False   5d7h
```

EtcdCluster conditions (lastTransitionTime a day older than the outage):

```
Available:   True   QuorumHealthy  "All members are ready"
Degraded:    False  QuorumHealthy
Progressing: False  Reconciled     "actual state matches status.observed"
readyMembers: 3
```

Every EtcdMember:

```
Ready: False  PodNotReady  "pod phase: Succeeded"
```

Operator log, one entry per cluster every ~15 s, nothing else:

```
ERROR  MemberList failed  {"controller": "etcdcluster", ..., "error": "context deadline exceeded"}
  ... dial tcp: lookup etcd-7p9fr.etcd.tenant-portal.svc on 10.96.0.10:53: no such host
github.com/cozystack/etcd-operator/controllers.(*EtcdClusterReconciler).promotePendingLearner
	controllers/etcdcluster_controller.go:1010
github.com/cozystack/etcd-operator/controllers.(*EtcdClusterReconciler).Reconcile
	controllers/etcdcluster_controller.go:344
```

## Code path (main @ a3d1b19)

`Reconcile`, steady-state branch (`current == desired`):

- `controllers/etcdcluster_controller.go:344` calls `promotePendingLearner`.
- `controllers/etcdcluster_controller.go:1008-1011`: `MemberList` fails, the error is logged, and the function returns `&ctrl.Result{RequeueAfter: 10s}, nil`.
- `controllers/etcdcluster_controller.go:348-350`: `if res != nil { return *res, nil }` returns from `Reconcile`.
- `controllers/etcdcluster_controller.go:379`: `updateStatus` is never reached.

`updateStatus` (`controllers/etcdcluster_controller.go:1334`) is the only place that recomputes `readyMembers` from the members' `Ready` conditions and derives `Available` / `Degraded` from that count. Had it run it would have written `0/3 members ready, quorum lost`. The PDB reconcile also lives there, so the PDB goes stale as well.

The sibling error branches directly above (TLS config, credentials, dial failure, lines 328-341) each carry a comment saying not to bail out of the whole reconcile because `updateStatus` still has useful work. The MemberList-failure branch does exactly what those comments warn against.

## Why it matters

Anything that gates on `Available` (Flux/Argo health checks, alerting, the tenant Kubernetes control plane in Cozystack, a human running `kubectl get etcdclusters`) is told the cluster is healthy precisely when it is not. The failure mode is silent: the operator is not crashing, the object is being reconciled every 15 s, and the status looks fresh.

## Things to think about before fixing

The one-line fix (return `nil, nil` from the MemberList-failure branch so the reconcile falls through to `updateStatus`) addresses this instance. Some broader questions are worth settling at the same time rather than after the next occurrence:

1. **Should status ever depend on etcd being reachable?** `updateStatus` only needs the EtcdMember list from the apiserver. Structurally, status derivation could run unconditionally (for example via `defer`, or by moving it ahead of every etcd-dialling step) so that no future early return can freeze it again. `Reconcile` has around a dozen `return &ctrl.Result{RequeueAfter: 10s}` sites in `scaleUp` / `completePendingMember` / `tryPromoteLearner` (lines 1144-1198) that would all skip status today.
2. **Is "all member pods Ready" the right definition of Available?** A member's `Ready` mirrors its Pod's readiness condition. A pod can be Ready while etcd has lost quorum (partition, disk stall) and the operator would still report `QuorumHealthy`. A MemberList/Status RPC failing across every endpoint is itself strong evidence that quorum is gone; it may deserve to feed `Available` directly (there is already an `Available=False / ClusterUnreachable` reason used during discovery at line 684).
3. **Should status carry a staleness signal?** Something like an observed timestamp or an `Unknown` condition status when the operator has not been able to verify the cluster for N reconciles, so consumers can distinguish "verified healthy" from "last known healthy".
4. **Test coverage.** An envtest/unit case where MemberList fails while members are not Ready, asserting `Available` flips to False, would have caught this and guards the fix.

Related: the members in this incident stayed down because their Pods, having reached a terminal phase, are never replaced (#368).

androndo:
## Problem

Fixes #367.

On a **converged** cluster (`ClusterID` latched, `current == desired`) the steady-state `promotePendingLearner` and `reconcileAuth` attempts returned their transient requeue directly, short-circuiting the reconcile **before** `updateStatus`. When etcd became unreachable the promote path hit its `MemberList`-failure branch (`RequeueAfter: 10s`) and returned early, so the cluster's `Available` / `Degraded` conditions and `ReadyMembers` **froze at their last-healthy value**.

Result: a fully down cluster kept advertising `Available=True` / `QuorumHealthy` indefinitely, even though every `EtcdMember` had already flipped `Ready=False`. Monitoring/GitOps health checks gating on the `Available` condition received a false-healthy signal.

## Fix

`updateStatus` already derives the cluster conditions purely from the `EtcdMember` set, so running it on every converged pass reports the truth on its own — no dependence on the operator's own etcd dial.

- Thread the promote/auth transient requeue through `updateStatus` via a new `pending *ctrl.Result` parameter instead of returning early.
- The status write always happens; the returned `Result` carries whichever requeue fires sooner (`soonerRequeue`) — the promote/auth 5–10s retry still wins over `updateStatus`'s 30s steady cadence.

This mirrors the **existing** "log and fall through to `updateStatus`" stance already applied to the TLS-config, credentials, and dial-failure branches in the very same block (see the comments there).

Genuine errors (`perr` from promote, `err` from auth) still bubble as before — only the transient no-error requeues are threaded.

## Test

`TestReconcile_UnreachableEtcdDoesNotFreezeStatus`: converged cluster, stale `Available=True/QuorumHealthy` in status, 3 members `Ready=False`, a dialable fake etcd whose `MemberList` errors. After reconcile it asserts `Available=False`, `ReadyMembers=0`, and `RequeueAfter=10s` (the promote requeue survived). Verified it **fails** against the pre-fix early-return and passes with the fix.

`make test` / `go test ./controllers/` green; `go vet` clean.

## Not in scope

Scope: this PR fixes the freeze on the **converged** path (`current == desired`, all pods down). The same freeze class on the scale and in-flight-member-deletion early returns is **not** addressed here and is tracked in #371 — it turns on the durable "derive member status on every non-error return" shape, which is worth settling before a second early-return patch (see the review thread and #371 for the progression-time quorum-denominator subtlety).

Also deliberately left out:
- A dedicated staleness/`Unknown` condition: the per-member `Ready=False` already carries the truth once `updateStatus` runs, and `Available=False/QuorumLost` is surfaced correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cluster status and availability information continue updating when etcd is temporarily unreachable.
  * Pending promotion and authentication retries no longer delay status or PodDisruptionBudget updates.
  * Retry timing now prioritizes immediate actions while preserving the shortest applicable requeue interval.
  * Clusters remain marked as progressing until pending retries complete.
  * Availability and degraded conditions now accurately reflect quorum loss and unhealthy established voters during scaling or member failures.
  * Availability and degraded status no longer fluctuate incorrectly during healthy scale-up.
  * PodDisruptionBudget protection now reflects the current voter count.


lllamnyp (changes requested):
Thanks for the quick turnaround on #367. The change does cure the exact case in the issue, CI is green, and the test fails against main as claimed. Requesting changes because threading the promote/auth requeue into `updateStatus` introduces two status-semantics regressions, leaves two sibling freezes of the same class in place, and the new helper has a latent bug. Details below, most severe first.

### 1. Degraded now flaps on every routine scale-up

`controllers/etcdcluster_controller.go:354`

When the last learner has been added, `current == desired`, so the promote block runs. A learner-not-ready retry (`tryPromoteLearner` → `RequeueAfter: 5s`) used to return early with no status write. It now falls into `updateStatus` with `ready == 2`, `desired == 3`, which hits the `ready > desired/2` branch and writes `Available=True/QuorumAvailable "2/3 members ready"` plus `Degraded=True/MembersUnhealthy`, while `Progressing=True`.

`docs/concepts.md` defines `Degraded` as the condition alerting pages on. The `!allMembersReady` gate at line 306-308 still suppresses the identical write for earlier learners in the same scale-up, so two identical mid-scale-up windows now report differently. No test asserts `Degraded` across a scale-up.

Suggested direction: keep threading `pending`, but do not rewrite the Available/Degraded pair while a not-ready promote retry is pending (or while `Progressing=True` and `!reconciliationComplete`).

### 2. `Progressing=False/Reconciled` is stamped while a learner is still unpromoted or auth-enable is still failing

`controllers/etcdcluster_controller.go:1461`

`reconciliationComplete` checks count, `ClusterID` and `MemberReady` only; `MemberReady` ignores `IsLearner`. On main the Reconciled write could only happen on the pass after promotion/latch succeeded, because the promote/auth requeue returned first. Now, with a learner whose `MemberPromote` keeps being rejected (raft log not caught up), or with `spec.auth.enabled` and a missing root Secret (`reconcileAuth` returns 10s forever), the cluster reads `Progressing=False/Reconciled`, `ProgressDeadline` is cleared, and `Available=True/QuorumHealthy "All members are ready"`, although etcd has 2 voters + 1 learner or auth was never applied. Anyone gating on `Progressing=False` proceeds early.

The root cause is `reconciliationComplete` ignoring `IsVoter` and `status.authEnabled`, which predates this PR, but this PR is what makes it observable. Either extend the predicate or keep the Reconciled write gated on `pending == nil`.

### 3. The freeze is not fixed for an in-flight EtcdMember deletion

`controllers/etcdcluster_controller.go:264`

The PR body calls this path "a bounded, local-CR transient". It is not bounded. `handleDeletion` (`controllers/etcdmember_controller.go:195-231`) never drops the finalizer until `removeMemberFromEtcd` returns nil and requeues 5s forever; with surviving peer CRs and etcd down, it errors forever. On the cluster side, `reconciliationComplete` is false (2 != 3), `deadlineExpired` is false because `ProgressDeadline` was nil'd at convergence (lines 1466-1469), and the deletion-wait loop returns 5s before `updateStatus`. `ReadyMembers=3 / Available=True/QuorumHealthy` freezes exactly as in #367. Scenario: break-glass member deletion (or the memory-backed self-delete on pod loss), then all pods go down.

### 4. The freeze is not fixed on any scale path

`controllers/etcdcluster_controller.go:308`

Every scale-related return (deletion wait 268, `hasPendingMember` → `scaleUp` 282, `!allMembersReady` 308, `scaleDown` 315, and the 5s/10s returns inside `scaleUp`/`completePendingMember`/`scaleDown`) skips `updateStatus`. Healthy 3/3, `replicas: 5`, member 4 added as learner, node shutdown takes all pods down: status advertises healthy with 0 ready pods for the full 600s deadline, then `handleDeadlineExceeded` writes `Available=False/DeadlineExceeded` and returns without touching `ReadyMembers` (stays 3), `Degraded` (stays False/QuorumHealthy) or the PDB, and every later pass re-enters it until the user edits spec.

Items 3 and 4 are the same class as #367. If they are out of scope here, the PR body should say so and a follow-up issue should track them; but together with 1 and 2 they suggest the durable shape is to derive the member-based status before the scale fan-out (or on every non-error return) and keep it separate from the reconcile-state conditions that other writers own (cert-manager gate at 121, `surfaceDiscoveryError` 651/678/694, `handleDeadlineExceeded` 2005/2025). That is the first open question on #367; worth deciding before landing a second early-return patch.

### 5. `soonerRequeue` drops a pending retry for a zero base

`controllers/etcdcluster_controller.go:1521`

`soonerRequeue(ctrl.Result{}, &ctrl.Result{RequeueAfter: 10s})` returns `ctrl.Result{}`: neither `Requeue` is set and `10s < 0` is false, so the transient retry is lost. Latent today because the only caller passes a 30s base, but the helper is package-level and documented as "whichever requeues sooner"; the next `soonerRequeue(ctrl.Result{}, pending)` caller silently loses its retry. Also, `{Requeue: true, RequeueAfter: 40s}` is treated as immediate, but controller-runtime v0.21 dispatches `RequeueAfter > 0` before `Requeue` (and `Result.Requeue` is deprecated there). Treat `base.RequeueAfter == 0 && !base.Requeue` as "never", or narrow the doc and signature to the one shape used, and add the zero-base case to `TestSoonerRequeue`.

### 6. The auth-gate comment gives the wrong reason

`controllers/etcdcluster_controller.go:372`

"auth-enable must not race an in-flight promotion": `tryPromoteLearner` calls `MemberPromote` synchronously under a 10s timeout and cancels before returning, so nothing is in flight by the time `pending` is set, and enabling auth alongside a Ready-but-unpromoted learner is harmless. What the gate actually protects against is a second etcd dial right after promote's `MemberList` timed out: `reconcileAuth`'s `AuthStatus`/`UserAdd`/`UserGrantRole`/`AuthEnable` use the raw reconcile ctx with no deadline, and clientv3 defaults to WaitForReady, so an ungated call against a dead endpoint stalls the worker. Please state that reason, or bound those calls with `context.WithTimeout` and drop the gate.

### 7. The `pending` parameter can be a caller-side merge

`controllers/etcdcluster_controller.go:1350`

`pending` is only read at the final return. `updateStatus`'s only error path returns `ctrl.Result{}, err`, so

```go
res, err := r.updateStatus(ctx, cluster, active)
if err != nil {
    return ctrl.Result{}, err
}
return soonerRequeue(res, pending), nil
```

at line 389 is equivalent, keeps the status writer decoupled from `Reconcile`'s control-flow state, and removes the `nil` at the other five call sites.

### Nits (non-blocking)

- `controllers/etcdcluster_controller_test.go:661`: `meta.FindStatusCondition(cluster.Status.Conditions, lll.ClusterAvailable)` instead of the hand-rolled loop (already used in `test/e2e/suite_test.go`).
- `controllers/etcdcluster_controller_test.go:614`: `metav1.Now().Add(time.Hour)` rather than `60 * 60 * 1e9`.
- `controllers/etcdcluster_controller_test.go:627`: the fixture block is a near-verbatim copy of five others in the file; a shared `members(t, n, readyStatus)` helper would let the test show only what it is about (members already `Ready=False`, stale `Available=True`).

androndo:
Thanks — pushed 73eea79 addressing items 1, 2, 5, 6, 7 and the nits; 3 and 4 deferred to #371 per your offer. Per-item:

**1 — Degraded flaps on scale-up.** `updateStatus` now withholds the Available/Degraded rewrite while `Progressing=True && !reconciliationComplete` (a new `case progressing:` in the health switch, computed via `clusterProgressing`). That's the same window the `!allMembersReady` early-return already suppresses for the earlier learners, so every mid-scale-up window now reports alike. The #367 down-cluster path is unaffected: there `Progressing=False` (Reconciled latched before the pods died), so the switch still writes `Available=False/QuorumLost`. New `TestUpdateStatus_NoDegradedFlapDuringScaleUp` pins it.

**2 — Reconciled stamped while a learner is unpromoted / auth failing.** Went with the localized option: the `Progressing=False/Reconciled` write (and the `ProgressDeadline` clear) is now gated on `pending == nil`. In both your scenarios `pending` is non-nil — `tryPromoteLearner` returns a 5s retry for the unpromoted learner, `reconcileAuth` a 10s retry for the missing Secret — so both are covered without widening `reconciliationComplete` (which also feeds spec-change adoption and could wedge future spec edits behind a stuck auth). New `TestUpdateStatus_WithholdsReconciledWhilePendingRetry` covers both the withheld and the settled pass.

**3 & 4 — deletion / scale-path freezes.** Deferred to #371. Agreed they're the same class as #367 and that the durable shape (derive member-status before the scale fan-out, decoupled from the reconcile-state conditions) is worth settling before a second early-return patch — especially the progression-time quorum-denominator question item 1 surfaces. #369 stays scoped to the converged path it fixes; the PR body now says so.

**5 — soonerRequeue zero base.** Rewrote ordering through a `requeueDelay` helper: a zero Result maps to `math.MaxInt64` (\"never\"), and `RequeueAfter>0` outranks the deprecated `Requeue` flag per controller-runtime v0.21. `soonerRequeue(ctrl.Result{}, pending)` now yields `pending`. `TestSoonerRequeue` gains the zero-base and `{Requeue:true, RequeueAfter:40s}` cases.

**6 — auth-gate comment.** Rewritten to the real reason: skipping avoids reconcileAuth's unbounded, WaitForReady clientv3 calls dialing a dead endpoint right after the promote pass found it unreachable — not a promotion race (promote is synchronous). Kept the gate rather than threading timeouts into the four auth RPCs.

**7 — pending as a caller-side merge.** Item 2's fix makes `updateStatus` read `pending` as a genuine status input (it gates the Reconciled stamp), not just a Result to forward at the final return, so I kept the parameter and documented that role on the function. Happy to revisit if you'd still prefer the split.

**Nits:** `meta.FindStatusCondition` in the freeze test; `metav1.Now().Add(time.Hour)`; and a shared `scaleUpMembers` fixture the freeze test and both new tests use.

lllamnyp (changes requested):
Thanks, 73eea79 addresses 2, 5, 6, 7 and the nits cleanly, and #371 is a good home for 3 and 4. One blocking item remains: the fix for item 1 reopens the #367 freeze for a class of clusters.

### The `progressing` withhold has no quorum floor

`controllers/etcdcluster_controller.go`, the new `case progressing:` in `updateStatus`.

`progressing := !paused && clusterProgressing(cluster) && !reconciliationComplete(cluster, running)`. On the converged path (`current == desired`) that reduces to "Progressing is True and some member is not Ready". It cannot tell "one learner is still joining" from "every pod is dead", and in both cases it leaves Available/Degraded at their last value.

Reproduced against 73eea79 with the new fixture: a converged 3-member cluster carrying `Progressing=True/SpecChanged` and `Available=True/QuorumHealthy`, `scaleUpMembers(t, "test", "ns", 0, 3)` (all three `Ready=False`), `updateStatus(..., &ctrl.Result{RequeueAfter: 10s})`. After the write:

```
ReadyMembers=0  Available=True/QuorumHealthy "All members are ready"  Degraded=False
```

That is the #367 shape again, now with `readyMembers` and `Available` contradicting each other on the same object.

When does a converged cluster carry `Progressing=True`? For one pass after any spec-change adoption, and indefinitely whenever `pending` never clears: `spec.auth.enabled` with a missing or wrong root Secret (`reconcileAuth` loops on 10s), or a learner whose `MemberPromote` keeps being rejected. Item 2's new gate is exactly what holds `Progressing=True` in those states, so a cluster with a broken auth Secret that then loses all its pods advertises `QuorumHealthy` forever.

Ask: keep the withhold for the joining-member window, but give it a floor so quorum loss is always written. Minimal version:

```go
case progressing && ready > 0:
    // Health left as-is; a learner is still joining and the live voters hold quorum.
```

so `ready == 0` falls through to the `QuorumLost` branch. Better, since the fixture already sets `IsVoter` alongside readiness: withhold only while the Ready voters still hold quorum among the live voter set (`readyVoters > voters/2`), and write `QuorumLost` otherwise. Either way, please add the all-down-while-progressing case as a test next to `TestUpdateStatus_NoDegradedFlapDuringScaleUp`; it is a few lines on the same fixture. Settling the denominator for the general case can stay in #371.

### Note, non-blocking

`clusterProgressing` reads the cluster's own `Progressing` condition to decide what to write next, which is the self-referential status pattern #186 is about (status should be reconstructible from the world, not from the previous status). Fine for this PR given the scope, but the #371 shape should derive "a member is joining" from the member set (count vs observed, IsVoter) rather than from a condition this same function wrote on the previous pass.

lllamnyp (approved):
Approving. 738f32e and 3b4e369 close the all-down hole, and I pushed 0c210ad on top for the last corner: with the floor keyed on quorum, an established voter losing its pod while Progressing is latched still read "All members are ready" with Degraded=False. The withhold is now keyed on every established voter being Ready (the only not-Ready members are joiners), so a down voter falls through to QuorumAvailable/MembersUnhealthy mid-progression; the all-down and scale-up-flap shapes behave as before. Test added for that shape. Merging once CI is green.

